PR #11, opened 2025-03-31
chore: upgrade to latest storacha/go-ucanto

volmedo:
The client was using the old `web3-storage/go-ucanto`. This PR upgrades the dependency to point to the latest tag of `storacha/go-ucanto` and adjusts the code to the breaking changes in the lib's API.


alanshaw (approved):
👌

PR #23, opened 2025-06-09
refactor: Simplify creating a `ClientConfig`

Peeja:
Consolidates this pattern into a single function.

#### PR Dependency Tree

* **PR #22**
  * **PR #23** 👈


volmedo:
hey! sorry for the post-review comment, but if I'm still in time I'd like to propose renaming the `ClientConfig` type to just `Config`. It is defined in the `client` package, so it will be consumed by users of the package as `client.Config`. If you like the idea and end up implementing it, don't forget to rename `NewClientConfig` to `NewConfig` too 😄.

PR #24, opened 2025-06-10
feat: Client can invoke access abilities

Peeja:


pkg/client/requestaccess.go
```diff
@@ -0,0 +1,59 @@
...
+var SpaceAccess = []string{
+	"space/*",
+	"blob/*",
+	"index/*",
+	"store/*",
+	"upload/*",
+	"access/*",
+	"filecoin/*",
+	"usage/*",
+}
```
volmedo, line 24: what about defining this directly as a slice of `access.CapabilityRequest`?
```suggestion
var SpaceAccess = []access.CapabilityRequest{
	{Can: "space/*"},
	{Can: "blob/*"},
	{Can: "index/*"},
	{Can: "store/*"},
	{Can: "upload/*"},
	{Can: "access/*"},
	{Can: "filecoin/*"},
	{Can: "usage/*"},
}
```
that way you can avoid converting it each time afterwards. Not as clean maybe, but still straightforward enough.

pkg/client/requestaccess_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+		nb, err := access.AuthorizeCaveatsReader.Read(capability.Nb())
+		require.NoError(t, err, "expected to read the caveats without error")
+		require.Equal(t, "did:mailto:example.com:alice", nb.Iss.String(), "expected to authorize the correct issuer")
+
+		requestedCapabilities := make([]string, 0, len(nb.Att))
+		for _, att := range nb.Att {
+			requestedCapabilities = append(requestedCapabilities, att.Can)
+		}
+		require.ElementsMatch(t, client.SpaceAccess, requestedCapabilities, "expected to authorize the correct capabilities")
```
volmedo, line 47: I think it may be desirable to make `client.SpaceAccess` unexported and hardcode the list again here in the test. That way, if we ever remove a required capability accidentally, the test will tell us. Otherwise we are just testing the function is using the variable, but not that it's definition is correct.

volmedo (approved):
LGTM, love the small, incremental PRs 🎉 

Left a couple minor comments.

pkg/client/requestaccess.go
Peeja: Yeah, I like that. This was me over-literally translating JS. 😄

pkg/client/requestaccess_test.go
Peeja: Yeah, I was on the fence about this. You've convinced me.

PR #25, opened 2025-06-11
feat: Client can claim access

### LINKED ISSUE #17: Implement access invocations
Ref:
https://github.com/storacha/upload-service/blob/main/packages/access-client/src/access.js

Peeja:
Uses a different approach to testing. I'll reconcile those in a future PR when there's more client implemented and I can be confident in the right direction.

Closes #17 

#### PR Dependency Tree

* **PR #25** 👈
  * **PR #27**
    * **PR #28**
      * **PR #29**
        * **PR #30**

### DIFF
--- a/pkg/client/claimaccess_test.go
+++ b/pkg/client/claimaccess_test.go
@@ -73,7 +73,7 @@ func TestClaimAccess(t *testing.T) {
 		require.Equal(t, storedDel.Link().String(), claimedDels[0].Link().String(), "expected the claimed delegation to match the stored one")
 	})
 
-	t.Run("invokes `access/claim`", func(t *testing.T) {
+	t.Run("returns any handler error", func(t *testing.T) {
 		agent := helpers.Must(ed25519.Generate())
 
 		service := helpers.Must(signer.Wrap(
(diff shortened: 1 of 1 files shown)


pkg/client/claimaccess.go
```diff
@@ -0,0 +1,84 @@
...
+
+	resp, err := uclient.Execute([]invocation.Invocation{inv}, cfg.conn)
+	if err != nil {
+		return nil, fmt.Errorf("sending invocation: %w", err)
+	}
+
+	reader, err := receipt.NewReceiptReaderFromTypes[access.ClaimOk, serverdatamodel.HandlerExecutionErrorModel](access.ClaimOkType(), serverdatamodel.HandlerExecutionErrorType(), captypes.Converters...)
+	if err != nil {
+		return nil, fmt.Errorf("generating receipt reader: %w", err)
+	}
```
volmedo, line 47: in https://github.com/storacha/go-libstoracha/pull/32 we added convenience type definitions for receipts and receipt readers for a number of capabilities. Feel free to add them there to be used here.

On a second look of that other PR, I think the usability can be improved further, but that is something we can iterate on when we use it more.

volmedo (approved):
LGTM, a couple non-blocking suggestions.

pkg/client/claimaccess.go
Peeja: I'm going to put this off for a bit, but this is super helpful to know.

volmedo (approved):

PR #28, opened 2025-06-13
refactor: s/complex mock connection/simple server/

Peeja:


pkg/client/requestaccess_test.go
```diff
@@ -4,40 +4,49 @@ import (
...
-		require.NoError(t, err)
+		agentPrincipal := uhelpers.Must(ed25519.Generate())
 
-		account, err := did.Parse("did:mailto:example.com:alice")
-		require.NoError(t, err)
+		account := uhelpers.Must(did.Parse("did:mailto:example.com:alice"))
 
-		servicePrincipal, err := did.Parse("did:web:storage.example.com")
-		require.NoError(t, err)
+		invokedCapabilities := make([]ucan.Capability[access.AuthorizeCaveats], 0)
```
volmedo, line 25: hint: when initializing a slice with 0 length, you can also do
```
invokedCapabilities := []ucan.Capability[access.AuthorizeCaveats]{}
```
I find it a little less weird than the `0` argument to make, but that is of course very subjective.

pkg/testing/helpers/serverconnection.go
```diff
@@ -0,0 +1,29 @@
...
+	uhelpers "github.com/storacha/go-ucanto/testing/helpers"
+)
+
+// NewServerConnection creates a new Ucanto server and a connection to it. It
+// accepts `server.Option`s to configure the server. This is mainly (if not
+// exclusively) to provide service methods.
+//
+// The server generates its own service principal. It has a `did:web:` DID for
+// realism and readability in errors and failures, but calling code should use
+// `server.ID()` to get it rather than assume knowledge of the DID it picks.
```
volmedo, line 18: should this say `connection.ID()` instead?
```suggestion
// `connection.ID()` to get it rather than assume knowledge of the DID it picks.
```

pkg/testing/helpers/serverconnection.go
volmedo, line 1: nitpick: I would not create a `testing/helpers` package just yet since they are only used in the client package tests so far. A `setupTestServer` function would work for now and would avoid having to use both a `helper` and a `uhelper`. The setup function can easily be moved to its own package if we need to use it in a different place in the future.

volmedo (approved):
I like the mock connection thing going away.

PR #32, opened 2025-06-17
feat: `w3 login` command

Peeja:
As of this PR:

* `login <email>` logs in as `<email>`.
* All other commands use the login proofs for their invocations.
* Commands can still take additional proofs on the command line, which will be used for that operation only.

There's now even more duplicated boilerplate from invocation to invocation, but the design to clean that up is for another PR.

#### PR Dependency Tree

* **PR #32** 👈


Peeja:
Right now, I can log in, and I see the delegations hit the file on disk, but when I then try to `ls` a space I own, I get:

```
2025/06/17 16:48:27 &{Name:0x140002047c0 Message:Claim {"can":"upload/list"} is not authorized
  - Capability {"can":"upload/list","with":"did:key:z6MkkNo1d4K8qtCAeMeGiRKfik446u57gRgTwEd6dLrRNsZg"} is not authorized because:
    - Capability can not be (self) issued by 'did:key:z6MkhtzgyUnT7oNqxBF6LECsVsrV8zqYFYrz4HzihiQWCDRS'
    - Capability can not be derived from prf:bafyreictgfrsomp35j6reo5bham4chh5twpei4mskurhert7iumnmwchpq because:
      - Unable to resolve 'did:web:staging.up.storacha.network' key Stack:0x140002047e0}
exit status 1
```

Which is confusing to me: this is coming from the server, which is `https://staging.up.storacha.network` (I've confirmed). Shouldn't it know the key for its own DID? Doesn't it have to to do anything? What am I missing?

Peeja:
@alanshaw [via Discord](https://discord.com/channels/1247475892435816553/1384636810780479540/1384815705584701541)

> I think staging was deployed from the legacy branch last time so doesn't understand did:web:staging.up.storacha.network as an audience yet.
>
> I'll deploy it from main today to fix.

PR #33, opened 2025-06-17
feat: Client can poll claim invocations

Peeja:


pkg/client/pollclaim.go
```diff
@@ -0,0 +1,91 @@
...
+)
+
+type tickChannelKey struct{}
+
+// WithTickChannel adds a tick channel to the context.
+func WithTickChannel(ctx context.Context, tickChan <-chan time.Time) context.Context {
+	return context.WithValue(ctx, tickChannelKey{}, tickChan)
+}
+
+func (c *Client) PollClaim(ctx context.Context, authOk access.AuthorizeOk) <-chan result.Result[[]delegation.Delegation, error] {
```
frrist, line 21: I like what you had prior to this change, accepting a timer directly. Please don't embed one in the context, it's an anti-pattern.

https://pkg.go.dev/context#pkg-overview
> Use context Values only for request-scoped data that transits processes and APIs, not for passing optional parameters to functions.

Peeja: I've read that, but I figured this was the kind of thing that belongs in the context. The problem with an argument is that then we need to pass it in everywhere, and in practice so far the only place we need to actually use it is a test. I could make a whole `Options` just for that, but that seems like overkill.

Is having the argument in the list and passing `nil` everywhere the Go way?

frrist: Based on the related test bits:
```go
responses = []result.Result[access.ClaimOk, error]{// three results}

// A channel of three ticks, ready to read
tickChan := make(chan time.Time, 3)
tickChan <- time.Now()
tickChan <- time.Now()
tickChan <- time.Now()

resultChan := c.PollClaim(client.WithTickChannel(testContext(t), tickChan),...)
```
looks like were creating a buffered channel, ticking it three times (a mock of time passing more-or-less), then calling the method. This way the ticker fires three times and we get back the expected results.

> Is having the argument in the list and passing nil everywhere the Go way?

It can be, for example the http Client:
```go
client := &http.Client{Transport: nil} // Uses default transport
```

But doing that here, in a function signature seems kinda gross. Functional options are nice, I don't think they are overkill if they aid in testing. Or you could pass a duration instead of a ticker, in tests the duration can be very short, and in prod it can be longer.

pkg/client/pollclaim.go
```diff
@@ -0,0 +1,91 @@
...
+	resultChan := make(chan result.Result[[]delegation.Delegation, error])
+
+	go func() {
```
hannahhoward, line 24: another suggestion to simplify here: I find that go is easiest to read when you separate synchronous from asynchronous code

```golang
func (c *Client) PollClaim(ctx context.Context, authOk access.AuthorizeOk) <-chan result.Result[[]delegation.Delegation, error] {
    c.PollClaimWithTicker(ctx, authOk, time.Tick(250 * time.Millisecond))
}

func (c *Client) PollClaimWithTicker(ctx context.Context, authOk access.AuthorizeOk, ticker <-chan time.Time) <-chan result.Result[[]delegation.Delegation, error] {
   	resultChan := make(chan result.Result[[]delegation.Delegation, error], 1)
	go func() {
	     resultChan <- result.Wrap(func() ([]delegation.Delegation, error) {
	         return c.pollClaimWithTicker(ctx, authOk, ticker)
	     })
	     close(result)
        }()					            
}

func (c *Client) pollClaimWithTicker(ctx context.Context, authOk access.AuthorizeOk, ticker <-chan time.Time) ([]delegation.Delegation, error) {
    for {
	  select {
		case <-ctx.Done():
			return nil, fmt.Errorf("context cancelled before delegations could be claimed: %w", ctx.Err()))
		case _, ok := <-tickChan:
			if !ok {
				return nil, fmt.Errorf("tickChan closed before delegations could be claimed: %w", ctx.Err()))
			}

			dels, err := c.ClaimAccess()
			if err != nil {
				fmt.Println("Failed to claim access:", err)
	                         return nil, fmt.Errorf("failed to claim access: %w", err))
			}

			// Collect all delegations whose "access/request" fact matches the request
			// link in authOk
			relevantDels := make([]delegation.Delegation, 0, len(dels))
			for _, del := range dels {
				for _, fact := range del.Facts() {
					requestLinkValue, ok := fact["access/request"]
					if !ok {
						continue // Skip if the fact does not contain "access/request"
					}
					requestLinkValueNode, ok := requestLinkValue.(ipld.Node)
					if !ok {
						continue // Skip if the fact is not a valid IPLD Node for some reason
                                          }
					requestLink, err := requestLinkValueNode.AsLink()
					if err != nil {
						continue // Skip if the "access/request" fact is not a link
					}
					if requestLink.String() == authOk.Request.String() {
						relevantDels = append(relevantDels, del)
						break // No need to check further facts for this delegation
					}
				}
			}

			if len(relevantDels) > 0 {
				return relevantDels, nil
			}
		}
	}
}
```

hannahhoward (changes requested):
Many suggestions, I don't think any of them are blocking other than I would like to not use context.Context for passing the timer

pkg/client/pollclaim.go
frrist: And fwiw, a _good_ place to use context value is for things like tracing, e.g. https://github.com/open-telemetry/opentelemetry-go/blob/main/propagation/trace_context.go#L39 - where the context value contains metadata related to the trace of a request/operation propagating through a system.

Peeja: @hannahhoward Yeah, that's much cleaner. Done!

pkg/client/pollclaim.go
```diff
@@ -0,0 +1,71 @@
...
+		defer close(resultChan)
+
+		for {
+			select {
+			case <-ctx.Done():
+				resultChan <- result.Error[[]delegation.Delegation](fmt.Errorf("context cancelled before delegations could be claimed: %w", ctx.Err()))
+				return
+			case _, ok := <-tickChan:
+				if !ok {
+					resultChan <- result.Error[[]delegation.Delegation](fmt.Errorf("tickChan closed before delegations could be claimed: %w", ctx.Err()))
```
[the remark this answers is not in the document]
Peeja: > using range to iterate over a channel, the loop will automatically exit when the channel is closed

Right, but you can't do that *and* check `ctx.Done()` 😢

Peeja (commented):
(Apparently I started a review when I replied)

pkg/client/pollclaim.go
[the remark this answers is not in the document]
frrist: True! (I didn't mean to imply this pattern could be used here.). 
There's a nice alternative using the generator pattern - have the producer handle context cancellation and close the channel:
```go
// Producer handles cancellation and closes channel
go func() {
    defer close(ch)
    
    for i := 1; i <= 5; i++ {
        // Do context aware work (HTTP request, DB query, etc.)
        result, err := doSomeWork(ctx, i)
        if err != nil {
            if errors.Is(err, context.Canceled) {
                return // Context cancelled, deferred close() runs
            }
            continue
        }
        
        // Send result
        select {
        case ch <- result:
        case <-ctx.Done():
            return // Context cancelled, deferred close() runs
        }
    }
}()

// Consumer gets clean range syntax, when `ch` closes the loop exits.
for value := range ch {
    fmt.Printf("Received: %d\n", value)
}
```

pkg/client/requestaccess_test.go
```diff
@@ -45,7 +45,7 @@ func TestRequestAccess(t *testing.T) {
 		capability := invokedCapabilities[0]
 
 		nb := uhelpers.Must(access.AuthorizeCaveatsReader.Read(capability.Nb()))
-		require.Equal(t, "did:mailto:example.com:alice", nb.Iss.String(), "expected to authorize the correct issuer")
+		require.Equal(t, "did:mailto:example.com:alice", *nb.Iss, "expected to authorize the correct issuer")
```
frrist, line 48: `require.EqualValue` may help here, I would expect we want to compare with the string?

hannahhoward (approved):
LGTM

PR #35, opened 2025-06-17
refactor: Migrate `urfave/cli` from v2 to v3

Peeja:
This makes using the context work better. It looks like `cCtx.Context` did work before, but the docs say this ["This allows handler functions to utilize context.Context for blocking/time-specific operations"](https://cli.urfave.org/migrate-v2-to-v3/#handler-function-signatures-changes), so I think there are probably some hitches.

I didn't actually notice we were using (or could even get away with using) `cCtx.Context` until I was going through the migration, but it was so straightforward I was already pretty much done, so it seems like we may as well.

#### PR Dependency Tree

* **PR #33**
  * **PR #34**
    * **PR #35** 👈

### DIFF
--- a/cmd/upload.go
+++ b/cmd/upload.go
@@ -26,49 +26,49 @@ import (
 	"github.com/storacha/guppy/cmd/util"
 	"github.com/storacha/guppy/pkg/car/sharding"
 	"github.com/storacha/guppy/pkg/client"
-	"github.com/urfave/cli/v2"
+	"github.com/urfave/cli/v3"
 )
 
 // upload handles file and directory uploads to Storacha
-func upload(cCtx *cli.Context) error {
+func upload(ctx context.Context, cmd *cli.Command) error {
 	signer := util.MustGetSigner()
 	conn := util.MustGetConnection()
-	space := util.MustParseDID(cCtx.String("space"))
-	proofs := []delegation.Delegation{util.MustGetProof(cCtx.String("proof"))}
+	space := util.MustParseDID(cmd.String("space"))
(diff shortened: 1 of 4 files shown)


frrist:
Yeah, strong preference for Cobra (the CLI pkg) initially, Viper (the configuration pkg) can easily be integrated later. 

Short summary of why I prefer the snake libraries:
- A flag can be 'persistent', meaning that this flag will be available to the command it's assigned to as well as every command under that command. For global flags, assign a flag as a persistent flag on the root. This is great for things like `--api`, `--log-level`, `--config`, and other flags we want available on all commands w/o needing to redefine them.
- Flag Groups for when you have different flags that **must** be provided together (e.g. if they provide the --username flag they MUST provide the --password flag as well); Cobra can enforce that requirement. Additionally, cobra can prevent different flags from being provided together if they represent mutually exclusive options (e.g. `--yaml` and `--json`)
- Enforcement of arguments - you can require a command accept no args, or a per-defined number of args, nice for the email flow here where email is a required single argument.
- Command suggestions based on "closeness": https://github.com/spf13/cobra/blob/main/site/content/user_guide.md#suggestions-when-unknown-command-happens

The full benefit can be realized once cobra is combined with viper, allowing flags to be bound to the config automatically, example here: https://github.com/spf13/cobra/blob/main/site/content/user_guide.md#create-rootcmd

If you want a more thorough example of this, see here: https://github.com/storacha/piri/pull/110/

Peeja:
Looks like `cCtx.Channel` is actually working fine in the other branch, so I'll close this, if we're looking at other libraries.

PR #47, opened 2025-06-20
refactor: Stop passing proofs through lots of calls

Peeja:
Instead, add them to the client at the top and let the client use its proofs whenever it makes an invocation.

On the CLI, continue to not save proofs specified at the command line to the persisted agent state (by not enabling persistence for that client instance).

#### PR Dependency Tree

* **PR #47** 👈

### DIFF
--- a/pkg/client/spaceblobadd_test.go
+++ b/pkg/client/spaceblobadd_test.go
@@ -47,7 +47,7 @@ import (
 	"github.com/storacha/guppy/pkg/client"
 )
 
-func TestBlobAdd(t *testing.T) {
+func TestSpaceBlobAdd(t *testing.T) {
 	space, err := ed25519signer.Generate()
 	if err != nil {
 		t.Fatal(err)
@@ -88,9 +88,11 @@ func TestBlobAdd(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	c.AddProofs(proof)
+
 	testBlob := bytes.NewReader([]byte("test"))
 
-	_, _, err = c.SpaceBlobAdd(testContext(t), testBlob, space.DID(), receiptsURL, proof)
+	_, _, err = c.SpaceBlobAdd(testContext(t), testBlob, space.DID(), receiptsURL)
 	require.NoError(t, err)
 }
 
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -170,7 +170,6 @@ func reset(cCtx *cli.Context) error {
 }
 
 func ls(cCtx *cli.Context) error {
-	c := util.MustGetClient()
 	space := util.MustParseDID(cCtx.String("space"))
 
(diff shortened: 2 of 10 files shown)


pkg/client/invokeandexecute.go
```diff
@@ -27,9 +27,13 @@ func invokeAndExecute[Caveats, Out any](
 	with ucan.Resource,
 	caveats Caveats,
 	successType schema.Type,
-	options ...delegation.Option,
```
volmedo, line 30: aren't there any other options we could ever be interested in passing here?

volmedo (commented):
I'm trying to evaluate how much of an issue it is that data is not saved when proofs are provided, but I'm not sure how this is used effectively.

Adding a new property to the client to store these "externally provided" proofs, separated from the ones in AgentData, could be an option to use these proofs without disabling saving AgentData. Would such an approach be worth it?

pkg/client/invokeandexecute.go
Peeja: Yeah, that seems fair. `invokeAndExecute` would take them, but not the exported client methods (which themselves might decide to create and pass some to `invokeAndExecute`. I'll add it back to the signature here.

pkg/client/invokeandexecute.go
```diff
@@ -29,7 +29,12 @@ func invokeAndExecute[Caveats, Out any](
 	successType schema.Type,
 	options ...delegation.Option,
 ) (result.Result[Out, failure.IPLDBuilderFailure], fx.Effects, error) {
-	inv, err := capParser.Invoke(c.Issuer(), c.Connection().ID(), with, caveats, options...)
+	pfs := make([]delegation.Proof, 0, len(c.Proofs()))
+	for _, del := range c.Proofs() {
+		pfs = append(pfs, delegation.FromDelegation(del))
+	}
+
+	inv, err := capParser.Invoke(c.Issuer(), c.Connection().ID(), with, caveats, append(options, delegation.WithProof(pfs...))...)
```
volmedo, line 37: very little nitpick:
```suggestion
	options = append(options, delegation.WithProof(pfs...))
	inv, err := capParser.Invoke(c.Issuer(), c.Connection().ID(), with, caveats, options...)
```

pkg/client/spaceblobadd_test.go
```diff
@@ -47,7 +47,7 @@ import (
 	"github.com/storacha/guppy/pkg/client"
 )
 
-func TestBlobAdd(t *testing.T) {
+func TestSpaceBlobAdd(t *testing.T) {
```
volmedo, line 50: ❤️

pkg/client/invokeandexecute.go
Peeja: Ah! 👍🏻

PR #52, opened 2025-06-24
refactor(uploads): reorganize uploads

hannahhoward:
# Goals

reorganizes uploads to match the storacha domain entity "Upload" which represents a single successful upload of a DAG

# Implementation

- where a scan = 1 filesystem scan, and upload equals as many filesystem scans, dag assembly attempts, and shard uploads as are needed to successfully upload the full dag from a source
- configuration = a set of parameters that apply to multiple sources and or uploads
- configuration_sources = connect sources to configurations, so I can say "set of uploads for every source attached to this configuration" -- not 100% sure if this is needed
- also adds a schema.sql

#### PR Dependency Tree

* **PR #52** 👈
  * **PR #53**
    * **PR #54**
      * **PR #57**
        * **PR #58**

### DIFF
--- a/pkg/preparation/sqlrepo/schema.sql
+++ b/pkg/preparation/sqlrepo/schema.sql
@@ -1,0 +1,122 @@
+-- enable foreign key constraints
+PRAGMA foreign_keys = ON;
+-- enable write ahead logging
+PRAGMA journal_mode = WAL;
+
+-- DROP TABLE IF EXISTS sources CASCADE;
+-- DROP TABLE IF EXISTS configurations CASCADE;
+-- DROP TABLE IF EXISTS configuration_sources;
+-- DROP TABLE IF EXISTS uploads CASCADE;
+-- DROP TABLE IF EXISTS scans CASCADE;
+-- DROP TABLE IF EXISTS fs_entries CASCADE;
+-- DROP TABLE IF EXISTS directory_children;
+-- DROP TABLE IF EXISTS dag_scans CASCADE;
+-- DROP TABLE IF EXISTS nodes CASCADE;
+-- DROP TABLE IF EXISTS links;
+CREATE TABLE IF NOT EXISTS sources (
+  id BLOB PRIMARY KEY,
+  name TEXT NOT NULL,
+  kind TEXT NOT NULL,
+  path TEXT NOT NULL,
+  connection_params BLOB,
+  created_at TEXT NOT NULL,
+  updated_at TEXT NOT NULL
+) STRICT;
+
+CREATE TABLE IF NOT EXISTS configurations (
+  id BLOB PRIMARY KEY,
+  name TEXT NOT NULL,
+  created_at TEXT NOT NULL,
+  block_size INTEGER NOT NULL,
+  links_per_node INTEGER NOT NULL,
+  shard_size INTEGER NOT NULL,
+  use_hamt_directory_size INTEGER NOT NULL,
+) STRICT;
+
+CREATE TABLE IF NOT EXISTS configuration_sources (
+  source_id BLOB NOT NULL,
+  configuration_id BLOB NOT NULL,
+  FOREIGN KEY (source_id) REFERENCES sources(id),
+  FOREIGN KEY (configuration_id) REFERENCES configurations(id),
+  PRIMARY KEY (source_id, configuration_id) 
+) STRICT;
+
+CREATE TABLE IF NOT EXISTS uploads (
+  id BLOB PRIMARY KEY,
+  configuration_id BLOB NOT NULL,
+  source_id BLOB NOT NULL,
+  created_at TEXT NOT NULL
+  FOREIGN KEY (configuration_id) REFERENCES configurations(id),
+  FOREIGN KEY (source_id) REFERENCES sources(id)
+) STRICT;
+
+CREATE TABLE IF NOT EXISTS scans (
+  id BLOB PRIMARY KEY,
+  upload_id BLOB NOT NULL,
+  created_at TEXT NOT NULL,
+  root_id BLOB,
+  updated_at TEXT NOT NULL,
+  error_message TEXT,
+  state TEXT NOT NULL
+  FOREIGN KEY (upload_id) REFERENCES uploads(id),
+  FOREIGN KEY (root_id) REFERENCES fs_entries(id)
+) STRICT;
+
+CREATE TABLE IF NOT EXISTS fs_entries (
+  id BLOB PRIMARY KEY,
+  source_id BLOB NOT NULL,
+  path TEXT NOT NULL,
+  last_modified TEXT NOT NULL,
+  mode INTEGER NOT NULL,
+  size INTEGER NOT NULL,
+  checksum BLOB,
+  FOREIGN KEY (source_id) REFERENCES sources(id)
+) STRICT;
+
+CREATE TABLE IF NOT EXISTS directory_children (
+  directory_id BLOB NOT NULL,
+  child_id BLOB NOT NULL,
+  FOREIGN KEY (directory_id) REFERENCES fs_entries(id),
+  FOREIGN KEY (child_id) REFERENCES fs_entries(id),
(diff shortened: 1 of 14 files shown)


Peeja (commented):
Some modeling questions, now that we've got a schema:

* An Upload belongs to a Configuration, which has and belongs to many Sources. An Upload also belongs to a Source. What are the Sources a Configuration has and belongs to? Are they different from those it "has and belongs to" by way of Uploads?
* A Node belongs to a Source, but it's really referring to (a particular range of) a particular file. Would it make more sense to point to the FS Entry, and drop `path` from `nodes`?
* Does an FS Entry need to have and belong to many children/parents? It should be a tree, not a DAG, no?
  * Or do we want to allow for a multitree, where we can (for instance) create a new root and have it contain the same children as the previous root?
* Should an FS Entry have a `name` instead of a `path`? I assume it's just one path segment, yeah?
* An FS Entry has many Scans, which belong to Uploads, which belong to Sources; so an FS Entry has many Sources through Scans and Uploads. But an FS Entry also *belongs to* one Source. How can that be?
  * We have a similar "problem" with FS Entries themselves: the FS Entry's children should all have the same Source as their parent(s).
  * If I'm reading correctly: multiple Scans can point at the same root FS Entry (specifically if the filesystem hasn't changed since the last Scan) but those scans must all be part of the same Upload—or at least must be part of Uploads which all have the same Source. Maybe there's no great structural way to express that constraint.

### Diagrams:
* [As is](https://dbdiagram.io/d/refactoruploads-reorganize-uploads-#52-685acc57f413ba3508a81f8a)
* [With `nodes` → `fs_entries` instead of `nodes` → `sources`](https://dbdiagram.io/d/refactoruploads-reorganize-uploads-#52-Suggestions-685acfc4f413ba3508a8bcf3)

pkg/preparation/sqlrepo/schema.sql
```diff
@@ -0,0 +1,122 @@
+-- enable foreign key constraints
+PRAGMA foreign_keys = ON;
```
frrist, line 2: https://stackoverflow.com/questions/64785806/why-is-sqlite-pragma-foreign-keys-not-persistent-across-db-connections worth a read, may need to set this on the connection level.

Feel free to crib from https://github.com/storacha/piri/tree/main/pkg/database/sqlitedb which attempts to provide a nice-ish interface for this.

hannahhoward:
Before I dig in, a general overview:

There's a key distinction here between a stateful, process entity, and immutable data entity (immutable in terms of state in the database). To understand it can be helpful to track these across the tree:

Process entities: Uploads, Scans, DAGScans
Immutable entities: Sources, FSEntry, directory_children (no model), Node, Link 

Process entities track something that happens and finishes (success/error/cancel).
Immutable entities track something that exists or existed at one time. (they are implicitly either valid or invalid, based of whether the thing they tracked still exists -- at some point we'll need to garbage collect the invalid stuff)

There's two sort of parallel trees of relationships happening through these, with Configurations as the root of all (what is a configuration is addressed below)

The immutable tree is essentially:

Configuration -> many Sources -> many FSEntries -> has and belongs many directory directory children
                                                      -> many Nodes (why not FSEntry is addressed below) -> has and belongs to many Links
                                                      
The mutable tree is:
Configuration -> many Uploads -> many Scans
                                                       -> many DAGScans

And then you have the confusing part, the connections:

Upload = process to track one complete upload from one source. Has as many scans as is needed to successfully upload. Setup to track the server entity of an upload -- i.e. you should get exactly 1 upload/add per upload
    - so Upload belongs to a source, the source being uploaded

Scan = process to track scannning of a source, generating FSEntiries and directory_children as needed
    - at the end, Scan "belongs to" an FSEntry, but that's just a way of saving the root of the directory tree at the end of the scan
    
DAGScan = process to track scanning a single FSEntry, generating Nodes and Links
   - at the end, has a CID representing a foreign key relationship with the root node for the FSEntry (cause cid is the primary key of a Node)

Ok so now the questions:
                                                      
> Some modeling questions, now that we've got a schema:
> 
> * An Upload belongs to a Configuration, which has and belongs to many Sources. An Upload also belongs to a Source. 

What are the Sources a Configuration has and belongs to? Are they different from those it "has and belongs to" by way of Uploads?

Yea Configurations are my least confident entity, and support pushback here. Let me run through:

The idea here is that this describes a series of sources you'd like to upload, that you'd like to upload 1 to N revisions of (with the >1 uploads being delta uploads), all sharing a specific confuguration for upload. The way I'd think of this is "I've got a dataset in 10 S3 buckets, I want to upload all 10 of them, then periodically update all ten of them over time". So I start with a Configuration with many sources. Then I say "let's make a new upload for each source in the configuration" (note the CreateUploads function) which generates and runs an upload for each source which belongs to the configuration. I've thought of other names: Dataset, Collection, ....

I will say based off what I know, "i've got n data sources I want to upload at once and keep updating periodically over time" is a real user scenario that comes up a lot in enterprise uplaod

> * A Node belongs to a Source, but it's really referring to (a particular range of) a particular file. Would it make more sense to point to the FS Entry, and drop `path` from `nodes`?

So, it's referring to an immutable RANGE of data within a particular file at path within a source.

FS Entry refers to a particularly revision of a file or directory

A large file can change without all of the 1MB ranges changing. If a file changes, we make a new FSEntry. We could make new RawNodes for every block in the file. But we'd probably prefer to only upload changes. So either we have to make another many to many, or we just copy the relevant attributes to the node. Ultimatlely, path + source + range + cid is enough to verify if the node is valid, it seems much more ideal for what we want to do, which is pull the data as we put it into an shard upload.

> * Does an FS Entry need to have and belong to many children/parents? It should be a tree, not a DAG, no?
>   
>   * Or do we want to allow for a multitree, where we can (for instance) create a new root and have it contain the same children as the previous root?

no, it should be a DAG for the reason above. we should have exactly 1 FS entry for a given file/dir with a given checksum. a directories checksum changes if any of the constituent files/folders change, but that doesn't mean we should be making entries for the files that didn't change just to associate it with the new parent.

- so if I have directory A, containing file B and file C:
1. new entities, first scan = FSEntry for file B, FSEntry for file C, FSEntry for directory A, relationship A to B, relationtion A to C
2. now file C changes (either for a seperate upload or during a later stage of the upload that prevents it from succeeding)
3. new entities, second scan = FSEntry for file `C, FSEntry for `A (changed cause C changed), relationship `A to B, relationship `A to C

(but FSEntry B does not change)

> * Should an FS Entry have a `name` instead of a `path`? I assume it's just one path segment, yeah?

No it's a path. I don't want to deal with having to trace back through the parents just to get a path.

> * An FS Entry has many Scans, which belong to Uploads, which belong to Sources; so an FS Entry has many Sources through Scans and Uploads. But an FS Entry also _belongs to_ one Source. How can that be?

No I don't see how you get there

The only connection between FSEntry and Scan is the root FSEntry ID, saved after a scan completes. Scans belong to uploads. Uploads have one source, it will always be the same source as is on all the child FSEntries. It's a denormalization, but mostly cause I don't want to have to walk back through the entire tree to get to the source id.

>   
>   * We have a similar "problem" with FS Entries themselves: the FS Entry's children should all have the same Source as their parent(s).

They do, what's the problem? That it's not enforced at the DB level? That's just not gonna be efficient given the tree nature to join all the way back to the source.

An FS Entry file revision at a given path in a given source. I don't see why it wouldn't belong to a source.

>   * If I'm reading correctly: multiple Scans can point at the same root FS Entry (specifically if the filesystem hasn't changed since the last Scan) but those scans must all be part of the same Upload—or at least must be part of Uploads which all have the same Source. Maybe there's no great structural way to express that constraint.
> 

Yes, this is true, but 
> ### Diagrams:
> * [As is](https://dbdiagram.io/d/refactoruploads-reorganize-uploads-#52-685acc57f413ba3508a81f8a)
> * [With `nodes` → `fs_entries` instead of `nodes` → `sources`](https://dbdiagram.io/d/refactoruploads-reorganize-uploads-#52-Suggestions-685acfc4f413ba3508a8bcf3)

PR #53, opened 2025-06-25
refactor: Break `sqlrepo` into files by repo interface

Peeja:
Breaks out the repo implementation by model.

Also removes the composed `Repo` interface, on the theory that we probably won't ever use it: no client code ever needs to abstractly get a thing that can serve as a repo for all of the models. Instead, `sqlrepo.New()` returns a `repo` concretely, on the theory that one should [accept interfaces, return structs](https://bryanftan.medium.com/accept-interfaces-return-structs-in-go-d4cab29a301b).

#### PR Dependency Tree

* **PR #52**
  * **PR #53** 👈
    * **PR #54**
      * **PR #57**
        * **PR #58**


pkg/preparation/sqlrepo/sqlrepo.go
```diff
@@ -32,468 +18,11 @@ func Null[T any](v *T) sql.Null[T] {
...
-	uploads.Repo
-	sources.Repo
-	scans.Repo
-	configurations.Repo
-	uploads.Repo
-}
-
 // New creates a new Repo instance with the given database connection.
-func New(db *sql.DB) Repo {
+func New(db *sql.DB) *repo {
```
hannahhoward, line 22: I think this might be my only disagreement.

As a general rule, I don't like returning private structs from public functions... but I honestly cannot remember why I feel this way! I would probably just make `repo` into `Repo` now that the interface type is gone.

hannahhoward: mind you you can have private fields in public structs

hannahhoward (approved):
LGTM though see non blocking comment

pkg/preparation/sqlrepo/sqlrepo.go
hannahhoward: oh wait, I remembered.

this is the side effect of private structs:

```golang
package a
struct a { b: string }
struct A { b: string }

func (a *a) B() string { return a.b }
func (a *A) B() string { return a.b }

func NewPrivateA(b: string) *a { return &a{b} }
func NewPublicA(b:string) *A { return &A{b} }
```

```golang
package c
import (
 "a"
 )
 
 func c() {
   // neither of these will work
   privateA := &a.a{b: "hello"}
   publicA := &a.A{b: "hello"} // though possibly &a.A{} will?
   // both of these will work
   privateA := NewPrivateA("hello")
   publicA := NewPublicA("hello")
   // both of these will work
   fmt.Println(privateA.B())
   fmt.Println(publicA.B())
 }
 
 // the main diff
 // will not work
 func doSomethingWIthPrivateA(privateA *a.a) string { return privateA.B() }
 // will work
 func doSomethingWIthPublicA(publicA *a.A) string { return publicA.B() }
 // but maybe you should be doing this anyway
 type A interface {
   B() string
 }
 func doSomethingWithA(a A) string { return a.B() }
```

Honestly, not a strong enough argument to make it public

PR #54, opened 2025-06-25
refactor: Test `sqlrepo.CreateScan()`

Peeja:
These should go faster now that I have a pattern. How does this look?

#### PR Dependency Tree

* **PR #52**
  * **PR #53**
    * **PR #54** 👈
      * **PR #57**
        * **PR #58**

This tree was auto-generated by [Charcoal](https://github.com/danerwilliams/charcoal)

PR #57, opened 2025-06-27
feat: `GetScanByID()`; test with roundtrip

Peeja:
#### PR Dependency Tree

* **PR #52**
  * **PR #53**
    * **PR #54**
      * **PR #57** 👈
        * **PR #58**


hannahhoward (approved):
LGTM though unless we actually have non-test code using this function, I'd assume not add it in a test file and leave out of the Repo. we will probably need the function at some point so probably no harm

PR #58, opened 2025-06-27
refactor: `TestUpdateScan()`

Peeja:


pkg/preparation/util/time.go
```diff
@@ -0,0 +1,30 @@
+package util
```
hannahhoward, line 1: if we're going to be mocking time, I'd prefer we use a full library like benjohnson/clock

also, do we really need this? cant we simply test UpdatedAt > CreatedAt -- golang clock gets down to the nanosecond even on Windows.

hannahhoward: honestly, I'd prefer not to mock time in the scans unless it feels super important

hannahhoward (changes requested):

pkg/preparation/util/time.go
hannahhoward: I'm not opposed to mocking time, if that's important. But let's please use a library like https://github.com/benbjohnson/clock (technically no-longer maintained, though it just works, and we can always fork it if we need to update)

PR #60, opened 2025-06-30
feat(dags): complete dag module

hannahhoward:
# Goals

Almost done with full execution setup (still working on upload logic)

# Implementation

- add unixfs directory code (needs work... a lot of copied code)
- add dags repo implementation
- add logic for processing all dag scans for an upload
- add logic to handle upload states
- split the visitor unixfsnode only and all unixfs
- cid scanner for db


Peeja (approved):
Pushed syntax and reference error fixes. With that, LGTM. I haven't dug too deep yet, but it's at least a big step in the right direction, and I'll be polishing it as I go through in future PRs anyhow.

PR #61, opened 2025-07-01
refactor(dags): simplify unixfs usage

hannahhoward:
# Goals

Get rid of the mass of copied code from go-unixfsnode

# Implementation

This requires a bit of hacking, and we give up some configurability, but I think its worth it so that a mass of unrelated code doesn't live in this repository. Also this repo is already complicated a little less configurability is probably fine.

The only thing I'm mildly worried about is the HAMT sharding size. In JS we hardcode at 1000 -- https://github.com/storacha/upload-service/blob/main/packages/upload-client/src/unixfs.js#L6 . Whereas in go it's set to total node size > 256KB (probably more than 1000) -- could result in slight divergence in cids uploaded with js vs go.


pkg/preparation/dag/visitor/linksystem.go
```diff
@@ -0,0 +1,163 @@
...
+// LinkSystem that visits nodes upon storage. The LinkSystem doesn't easily do
+// this, because it wants to break up encoding, hashing, and storage into
+// separate steps and only let you configure each step separately. Here though,
+// we kind of need to do something custom for the entire thing. The only
+// available solution was to take the encoding step and do everything there --
+// encoding, hashing, and visiting steps -- then eat the cost of IPLD's
+// separate hashing + storage steps by making them as close as possible to
+// no-ops. This is a bit of a hack, but it enables us to use the default code
+// in https://github.com/ipfs/go-unixfsnode/tree/main/data/builder which keeps
+// a bunch of confusing complexity out of this codebase.
```
alanshaw, line 29: Could we fork and change it so it works in the way we're trying to hold it?

alanshaw (approved):

PR #66, opened 2025-07-01
feat: CLI command to run preparation

Peeja:
Very sketchy so far, really just a forcing function to help me put everything together. Feedback welcome, though.

`large-upload` is kind of a silly name, but I didn't want to spend much time on it yet. I suppose eventually this will supersede (or at least become an option on) `upload`.

#### PR Dependency Tree

* **PR #66** 👈
  * **PR #78**
    * **PR #79**
      * **PR #80**
        * **PR #83**
          * **PR #84**
            * **PR #82**
            * **PR #87**

This tree was auto-generated by [Charcoal](https://github.com/danerwilliams/charcoal)

PR #67, opened 2025-07-02
feat(uploads): implement uploads execution

hannahhoward:
# Goals

implements uploads execution as a whole including worker management, with placeholders for remaining steps

# Implementation

- Implement a upload execution function that can perform restarts, essentially via a state transition loop that resets on error until we reach a terminal state (including if we max out restarts).
- Implement an internal executor to manage actually starting and stopping workers for each upload attempt

# For Discussion

I image in #66 we will actually want to replace the direct call to "scans.ExecuteScan" with a call to uploads.ExecuteUpload

That does require:
- initializing the scans module, and then setting up a function that uses scans.ExecuteScan to implement the RunNewScan function required by the uploads module
- initializing the Dags module first, passing the worker to the uploads module


pkg/preparation/uploads/uploads.go
```diff
@@ -32,10 +40,244 @@ func (u Uploads) CreateUploads(ctx context.Context, configurationID types.Config
...
 }
 
 // GetUploadByID retrieves an upload by its unique ID.
 func (u Uploads) GetUploadByID(ctx context.Context, uploadID types.UploadID) (*model.Upload, error) {
-	return u.repo.GetUploadByID(ctx, uploadID)
+	return u.Repo.GetUploadByID(ctx, uploadID)
+}
+
+// ExecuteUpload executes the upload process for a given upload, handling its state transitions and processing steps.
+func (u Uploads) ExecuteUpload(ctx context.Context, upload *model.Upload, opts ...ExecutionOption) error {
```
Peeja, line 52: **question:** When an upload resumes from saved state in, doesn't it need to go back to the first step regardless of its current persisted `State()`? Or is that what it does? It looks to me like it starts from the step it left off on.

pkg/preparation/uploads/uploads.go
```diff
@@ -32,10 +40,244 @@ func (u Uploads) CreateUploads(ctx context.Context, configurationID types.Config
...
+			e.start()
+		case model.UploadStateScanning:
+			fsEntryID, err := u.RunNewScan(ctx, upload.ID(), func(id types.FSEntryID, isDirectory bool) error {
+				err := u.Repo.CreateDAGScan(ctx, id, isDirectory, upload.ID())
+				if err != nil {
+					return fmt.Errorf("creating DAG scan: %w", err)
+				}
+				// indicate new work is available for the DAG scan worker
+				select {
+				case e.dagWork <- struct{}{}:
```
Peeja, line 80: **question/suggestion:** Would the logic be simpler if we sent the work itself to the Scan Worker? We'd still back it with the DB, but we'd use it less as a mailbox, and solely as a record of what we've already done and can skip.

hannahhoward:
@Peeja is this good to merge? (at least to move things forward, it needs an iteration of cleanup)

PR #79, opened 2025-07-08
refactor: Use custom `ID` type, avoid `[:]`s

Peeja:


hannahhoward (approved):
LGTM -- net net, I'm on board. I'm surprised we don't use any other methods on uuid.UUID. You may want to add a .String() method here at some point for log statements and others.

PR #80, opened 2025-07-08
refactor: Use a custom `Timestamp` type

Peeja:
#### PR Dependency Tree

* **PR #80** 👈


Peeja:
Well, it's not just about SQL read and write, it's also about having a consistent domain layer. But the timestamps aren't terribly significant; I'm happy to skip this. I'll close it.

PR #81, opened 2025-07-17
feat(dags): build a caching node reader

hannahhoward:
# Goals

This is just a draft, and probably needs a little more analysis, but here is a basic function that will read the data for nodes and verify it

# Implementation

Implement a node reader with an lrucache you can optionally use, that optionally also checks hashes on data read


Peeja:
This is an old PR and incorporated already.

PR #82, opened 2025-07-17
feat: Preparation creates shards

Peeja:
Note that the last shard remains open at the end. It should be closed when the work ends; that will happen next, in #88.

May be worth reviewing by commit. There are some commits that could have been their own small PRs, but it was too tangled to pull them out from the middle of the history.

#### PR Dependency Tree

* **PR #82** 👈
  * **PR #89**
    * **PR #88**
      * **PR #90**
        * **PR #91**
          * **PR #92**

### DIFF
--- a/pkg/preparation/sqlrepo/shards.go
+++ b/pkg/preparation/sqlrepo/shards.go
@@ -5,17 +5,12 @@ import (
 	"fmt"
 
 	"github.com/ipfs/go-cid"
-	cidlink "github.com/ipld/go-ipld-prime/linking/cid"
-	"github.com/multiformats/go-varint"
 	dagsmodel "github.com/storacha/guppy/pkg/preparation/dags/model"
 	"github.com/storacha/guppy/pkg/preparation/shards"
 	"github.com/storacha/guppy/pkg/preparation/shards/model"
 	"github.com/storacha/guppy/pkg/preparation/sqlrepo/util"
 	"github.com/storacha/guppy/pkg/preparation/types/id"
 )
-
-// Byte length of a CBOR encoded CAR header with zero roots.
-const noRootsHeaderLen = 17
 
 var _ shards.Repo = (*repo)(nil)
 
@@ -121,7 +116,7 @@ func (r *repo) FindNodeByCid(ctx context.Context, c cid.Cid) (dagsmodel.Node, er
 	return r.getNodeFromRow(row)
 }
 
-func (r *repo) CurrentSizeOfShard(ctx context.Context, shardID id.ShardID) (uint64, error) {
+func (r *repo) ForEachNodeCIDAndSize(ctx context.Context, shardID id.ShardID, yield func(cid.Cid, uint64)) error {
 	rows, err := r.db.QueryContext(ctx, `
 		SELECT nodes.cid, nodes.size
 		FROM nodes_in_shards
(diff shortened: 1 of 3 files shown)


pkg/preparation/dags/model/dagscan.go
```diff
@@ -64,7 +67,9 @@ type dagScan struct {
 	updatedAt    time.Time
 	errorMessage *string
 	state        DAGScanState
-	cid          *cid.Cid // rootID is the ID of the root directory of the scan, if it has been completed
+	// cid is the root CID of the DAG, if the scan has been completed, and
```
hannahhoward, line 70: concur with making optional cids through cid.Undef instead of *cid.Cid

pkg/preparation/sources/sources.go
```diff
@@ -23,13 +22,19 @@ func (e ErrUnrecognizedSourceKind) Error() string {
 // API is the API for accessing and managing sources.
 type API struct {
 	Repo Repo
+	// GetLocalFSForPathFn is called to get an fs.FS for a [model.LocalSourceKind] source.
```
hannahhoward, line 25: curious why not just use os.DirFS?

nb, but best I can tell this is used for testing. I would argue that afero IS a different source kind, and actually should be defined as such. That would mean instead of having a seperate depedency, you'd just have another case statement. could be called model.TestSourceKind or something to that effect.

hannahhoward: maybe that is harder to implement than it looks

pkg/preparation/shards/model/shard.go
```diff
@@ -0,0 +1,114 @@
...
+type ShardState string
+
+const (
+	// ShardStateOpen indicates that a shard is still accepting new nodes.
+	ShardStateOpen ShardState = "open"
+	// ShardStateClosed indicates that a shard is no longer accepting nodes, but
+	// is not yet added to the space.
+	ShardStateClosed ShardState = "closed"
+	// ShardStateAdded indicates that a shard has been added to the space.
+	ShardStateAdded ShardState = "added"
```
hannahhoward, line 22: are we going to have a ShardStateError? or just re-open a shard if some piece of data is not read correctly on upload

just curious

pkg/preparation/shards/model/shard.go
```diff
@@ -0,0 +1,114 @@
...
+		return true
+	default:
+		return false
+	}
+}
+
+type Shard struct {
+	id       id.ShardID
+	uploadID id.UploadID
+	cid      *cid.Cid
```
hannahhoward, line 37: why *cid.Cid here and cid.Undef on dagScan?

pkg/preparation/sqlrepo/shards.go
```diff
@@ -0,0 +1,176 @@
...
+		nodeCID.Bytes(),
+		shardID,
+	)
+	if err != nil {
+		return fmt.Errorf("failed to add node %s to shard %s: %w", nodeCID, shardID, err)
+	}
+	return nil
+}
+
+func (r *repo) RoomInShard(ctx context.Context, shard *model.Shard, nodeCID cid.Cid, config *configmodel.Configuration) (bool, error) {
```
hannahhoward, line 104: I think this is the only method that I kind of think there should be a refactor on.

You're putting domain logic (how shards get sized) into the DB repo interface.

I would rather see a method like:

```golang

```golang

type ShardNode interface {
   Cid() cid.Cid
   Size() uint64
}

func (r *Repo) IterateShardNodes(ctx context.Context, shard *model.Shard, cb func(ShardNode) error) error {
...
}
```

and then the rest of this logic (i.e. RoomInShard calculation) goes up to the Shards module.

I think this is my only blocker so far.

pkg/preparation/sqlrepo/uploads.go
```diff
@@ -166,14 +161,7 @@ func (r *repo) CIDForFSEntry(ctx context.Context, fsEntryID id.FSEntryID) (cid.C
 		return cid.Undef, err
 	}
 	if ds.State() != dagmodel.DAGScanStateCompleted {
-		switch ds.State() {
```
hannahhoward, line 169: I can't remember why I thought these should be nil error for Failed & Canceled, but should they be nil?

pkg/preparation/sqlrepo/util/util.go
```diff
@@ -30,20 +31,32 @@ func (ts tsScanner) Scan(value any) error {
...
-// cidScanner returns a sql.Scanner that scans a CID from a byte slice into the
-type cidScanner struct {
-	dst *cid.Cid
+func DbCid(cid *cid.Cid) dbCid {
+	return dbCid{cid: cid}
 }
 
-var _ sql.Scanner = cidScanner{}
+// dbCid returns a sql.Scanner that scans a CID from a byte slice into the
+type dbCid struct {
```
hannahhoward, line 43: this is great

pkg/preparation/types/id/id.go
```diff
@@ -31,6 +31,10 @@ func (id *ID) Scan(src any) error {
 	return nil
 }
 
+func (id ID) String() string {
```
hannahhoward, line 34: ALWAYS a good decision

hannahhoward (approved):
This is a really great PR @Peeja 

Overall LGTM with one required refactor -- I'd like to get the RoomInShard calculation out of the DB logic, since it really isn't specific to a database.

I put this as blocking but it can actually be fast follow if you want to merge

pkg/preparation/shards/model/shard.go
Peeja: Uh, because I changed the other one and forgot about this one? 😅 I'll make this one match.

pkg/preparation/shards/model/shard.go
Peeja: Not totally sure yet, but it should fall out from the next stage.

PR #90, opened 2025-07-31
refactor: Only signal next stage on new shards

Peeja:


hannahhoward (approved):
LGTM. This is the way :P

PR #93, opened 2025-08-04
feat: add receipt API client

alanshaw:
Converts the `pollAccept` function into a general receipts API client with polling.

Depends on https://github.com/storacha/go-libstoracha/pull/40


pkg/receipt/client.go
```diff
@@ -0,0 +1,159 @@
...
+
+// WithInterval configures the time to wait between poll requests. The default
+// is [PollInterval].
+func WithInterval(interval time.Duration) PollOption {
+	return func(opt *pollConfig) {
+		opt.interval = &interval
+	}
+}
+
+// WithRetries configures the maximum number of times that Poll will attenpt to
```
Peeja, line 113:
```suggestion
// WithRetries configures the maximum number of times that Poll will attempt to
```

pkg/receipt/client_test.go
```diff
@@ -0,0 +1,214 @@
+package receipt
```
Peeja, line 1: **suggestion:**


```suggestion
package receipt_test
```

We've been using separate test packages, and I don't _think_ there's a need not to do that here, but maybe I missed it.

PR #104, opened 2025-08-08
refactor: `NewPutClient` and family are `testutil`s

Peeja:


hannahhoward (approved):
Seems reasonable -- is it reused?

PR #103, opened 2025-08-08
refactor: `TestSpaceBlobAdd` checks put blob

Peeja:

PR #106, opened 2025-08-08
feat: `space/blob/add` shards as they close

Peeja:
This "adds" the shards to the space, but produces dummy data for the shards. The next PR will generate an actual shard CAR for each one.

#### PR Dependency Tree

* **PR #100**
  * **PR #101**
    * **PR #102**
      * **PR #103**
        * **PR #104**
          * **PR #105**
            * **PR #106** 👈
              * **PR #109**


hannahhoward (approved):
This great. I think you're going to find it may get a bit complicated once the errors are introduced but maybe I don't know.

Issue #41, opened 2025-06-20
Generate Indexes From Shards


hannahhoward:
- for each shard uploaded
- we will write records of block positions
- collect several shards up to a cid limit of some sort
- write out a index manifest

PR #108, opened 2025-08-12
feat:  Generate Indexes from Shards

Goddhi:
index generation feature from issue [#41](https://github.com/storacha/guppy/issues/41). 
- Track block positions within shards (offset and size)
- Generate JSON index manifests for completed uploads
- Store index manifests in database
- Automatic index generation when shards are closed


Goddhi:
Oh i opened PR shortly after you created the PR, seems you have implemented it already. i will close the PR right away. Thanks

PR #112, opened 2025-08-15
feat: After an error, can be retried safely

Peeja:
Big changes here!

* Adjusted the workers/APIs/domains to be much more consistent.
* Removed most of the upload states. An upload is now either pending or started. If an upload is started, it's failed if it has an error message, and it's successful if it has a `rootCID`. The state of the file scans, DAG scans, and shards are in those models.

Now, if the process fails at any point, the data remains frozen in the intermediate state it failed in, and can be restarted from there.

Notably, this does *not* directly address the underlying data on disk changing during the process, but that should also be easier to deal with now, as we can simply delete facts from the DB that we know are no longer true and idempotently throw the whole process at it again. But doing that is still for a future PR.

#### PR Dependency Tree

* **PR #112** 👈
  * **PR #115**
    * **PR #116**
      * **PR #117**
        * **PR #118**
          * **PR #124**
            * **PR #125**
              * **PR #126**
                * **PR #127**
                  * **PR #128**

### DIFF
--- a/pkg/preparation/uploads/uploads.go
+++ b/pkg/preparation/uploads/uploads.go
@@ -153,7 +153,7 @@ func (e *executor) runStartWorker(ctx context.Context, uploadAvailable <-chan st
 
 		// doWork
 		func() error {
-			if e.upload.NeedsStart() {
+			if e.upload.State() == model.UploadStatePending {
 				if err := e.upload.Start(); err != nil {
 					return fmt.Errorf("starting upload: %w", err)
 				}
--- a/pkg/preparation/uploads/uploads.go
+++ b/pkg/preparation/uploads/uploads.go
@@ -9,6 +9,7 @@ import (
 	logging "github.com/ipfs/go-log/v2"
 	"github.com/storacha/go-ucanto/did"
 	dagmodel "github.com/storacha/guppy/pkg/preparation/dags/model"
+	scanmodel "github.com/storacha/guppy/pkg/preparation/scans/model"
 	"github.com/storacha/guppy/pkg/preparation/types/id"
 	"github.com/storacha/guppy/pkg/preparation/uploads/model"
 	"golang.org/x/sync/errgroup"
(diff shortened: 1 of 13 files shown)


Peeja:
Amended last commit with SQL syntax fix. Whoops!

pkg/preparation/uploads/uploads.go
```diff
@@ -152,66 +143,95 @@ func (e executor) execute(ctx context.Context) (cid.Cid, error) {
...
+		// doWork
+		func() error {
+			if e.upload.State() == model.UploadStatePending {
+				if err := e.upload.Start(); err != nil {
+					return fmt.Errorf("starting upload: %w", err)
+				}
+				if err := e.api.Repo.UpdateUpload(ctx, e.upload); err != nil {
+					return fmt.Errorf("updating upload: %w", err)
+				}
+				if _, err := e.api.Repo.CreateScan(ctx, e.upload.ID()); err != nil {
```
hannahhoward, line 162: vaguely expect to update the upload AFTER the scan is created? Not sure why though.

hannahhoward: I guess ideal would be to transactionalize the whole thing but... meh

hannahhoward (changes requested):
- I don't think it's safe to delete RestartDAGScansForUpload
   - State transition is:
       -> Pending -> Running
       -> Running -> Complete or Cancelled
       -> The reason for restart is if there is a cancellation or process shutoff -- where the dag scan ends up in cancelled or or running state but without having finished, that should be reset and run again

-> The new scans worker also will not work without a restart command -- if the command is cancelled, likely a context cancellation will occur, so the scan will be in a cancelled state, and will not run again.

-> I don't think it's wise to remove the complete/cancelled states from upload. Instead, I would focus on the idea of a restart to put things back in a running state
   -> there are actually different behaviors to understand about a cancel vs a fail
      -> let's say the underlying disk data changes and that's the source of failure. If this happens, we need to trigger a NEW FS scan, even if the existing one is completed.
      -> In the case of a cancel, we no there was no actual error, and we can just resume
   -> How do I know if the error message is up to date? It doesn't appear to ever get cleared?  So if I'm re-running my scan, is the error current? or old?
   
Broadly, I agree with the part of the PR that adds the working for fs scans.

But:
- don't remove the restart logic from dag scans
- needs restart logic for fs scans
- fundamentally I just don't understand why we're getting rid of all the state logic for uploads -- it makes the state of the upload, which is intended to be very strict, much less clear.

The last point is one I'd like to discuss cause I feel like I'm just not seeing something.

Peeja:
@hannahhoward Yeah, I'm seeing it now. Let's nail down the specific situations that aren't handled right now and I can write tests for them so we know when we've covered them correctly. The `preparation_test`s are perfect for this.

PR #121, opened 2025-09-15
feat: standardize CLI output formatting

### LINKED ISSUE #120: Standardize CLI Output
Right now, the CLI shows output in different ways depending on the command. For example:

* `whoami` just prints the DID
* `login` shows a success message with an exclamation mark
* `upload` uses a symbol (⁂) before the URL
* Errors are also inconsistent (some plain, some wrapped with extra text)

This makes the CLI harder to use, especially when scripting or parsing the output.

## What We Should Do

* Use the same format for all success messages (e.g., `guppy: <message>`)
* Use consistent prefixes for errors (`error:`) and warnings (`warning:`)
* Add a global `--json` flag so every command can return structured output
* Create a small output helper so all commands share the same formatting

pranavkonde:
Fixes #120 

CLI output was inconsistent across commands, making it hard to parse and confusing for users. This PR introduces a standardized format for success messages, errors, and data display, with optional JSON output.

### Changes

* Added `internal/output` package with `Success()`, `Error()`, `Warning()`, `JSON()`, and `Table()` helpers
* Added global `--json` flag for all commands
* Unified error handling
* Improved command outputs (`whoami`, `login`, `reset`, `ls`) with consistent formatting and table display

### DIFF
--- a/guppy.go
+++ b/guppy.go
@@ -14,6 +14,7 @@ import (
 	"github.com/storacha/go-ucanto/core/delegation"
 	"github.com/storacha/go-ucanto/core/result"
 	"github.com/storacha/guppy/internal/cmdutil"
+	"github.com/storacha/guppy/internal/output"
 	"github.com/storacha/guppy/internal/upload"
 	"github.com/storacha/guppy/pkg/didmailto"
 	"github.com/urfave/cli/v2"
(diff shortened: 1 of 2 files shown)


hannahhoward:
I don't think this makes sense with where things have ended up

PR #126, opened 2025-09-17
feat: Upload UI, and demo

Peeja:
Presenting…Guppy!

success

#### PR Dependency Tree

* **PR #126** 👈
  * **PR #127**
    * **PR #128**


pkg/preparation/sqlrepo/stats.go
```diff
@@ -0,0 +1,101 @@
...
+import (
+	"context"
+
+	"github.com/storacha/guppy/pkg/preparation/types/id"
+)
+
+func (r *Repo) TotalBytesToScan(ctx context.Context, uploadID id.UploadID) (uint64, error) {
+	row := r.db.QueryRowContext(
+		ctx, `
+			SELECT COALESCE(SUM(fs_entries.size), 0)
```
hannahhoward, line 12: mild anxiety about the perf on this query at scale -- we could have maybe 100,000+ files in some uploads

hannahhoward (approved):
First of all, love the UI -- it's great.

Second, this is approved with some follow-ups.

1. I really don't think the demo should be part of the main command, and definitely I don't think demo files should live right next to non demo files (or in some cases, be part of the same file) -- I'm thinking maybe there should be a seperate /cmd/demo.

Seperately, we should move the command itself to /cmd (though maybe the old command is still there and we need to merge?)

Anyway, all a fast follow up but worth considering

2.  Hopefully the UI via a poll method reading from the database will scale. BUT I suspect we may need more of a push style callback that avoids additional DB queries from the upload process at some point. For now, let's just monitor

PR #130, opened 2025-09-22
refactor: Remove DAG scan state

Peeja:
Looks like we can get rid of this now. We have all the state we need in the presence (or not) of the CID, and we don't need to store error messages at this level, as they bubble out anyhow.


hannahhoward (changes requested):
so:
- I'm fine with getting ride of computed state that's actually a duplicate of other states
- I'm not fine with getting rid of data that contains actual new information that may change logic
- I believe getting rid of errors on the db will change logic

Retrying an upload after a read error while scanning a file will, unless I'm reading this wrong, reattempt to scan the same bad file, because there's no indicator that this FSEntry is no good. Moreover, it will attempt to re-dagscan all directory parents of that file, cause again, they aren't marked as no good.

I also believe actually this will prevent upload from ever completing, I think. Even on retry.

Long and short, I'm ok with getting rid of the state var but an FSEntry should only ever be run once (once you get past AwaitingChildren state) and the lack of a stored error means it may keep getting retried. maybe I'm missing something.

hannahhoward:
one way I would approach these is to ask yourself "could I actually produce the State() function from the remaining fields".

Awaiting children is a bit weird cause its a function of state stored on related models, but it's still there
Canceled doesn't mean anything -- you can just retry
Pending is just no cid 
Running is... not relevant at least for now, though we will have issues if we want to parallelize the operations here
Completed is just having a cid
Cancelled is effectively the same as pending
Failed is .... not something I can determine, unless there is a stored error -- and here's where I think it is important, because we don't want failed scans to run again.

hannahhoward (approved):

Peeja:
@hannahhoward Somehow your comment overwrote mine? I'm restoring mine to keep the flow, here's your reply:

---

> If the data has actually changed, the original data won't be scannable, but then this DAG scan won't attempt to rerun, because it won't show up in the FS scan.

Gonna run through the sequence of events as I see them, to make sure I'm not missing something:
- I start an upload, and do an FS Scan. In this scan, I encounter file `A`, make FSEntry `FS` for `A`, and then DAGScan `DS` for FSEntry `FS`
- By the time I get to running the DAGScan `DS`, someone has changed the file `A`. As such the checksum fails when running `DS` and DS does not generate a CID (but currently nothing is marked failed here). The error callback aborts the uplaod.
- I restart the upload, which does two things:
   - kick off a new FS Scan
   - startup the DAGScan worker, which immediately calls IncompleteDAGScansForUpload and begins processing IncompleteDAGScans
- The FS Scan will generate an FSEntry `FS′` for the new version of file `A`, and DAGScan `DS′` for FSEntry `FS′` -- DAGScan `DS′` will succeed now, which is great
- At the same time, the original DAGScan `DS` will get picked up by IncompleteDAGScansForUpload and run again, and fail again cause the checksum still doesn't match. This will happen every time the upload is restarted. This doesn't seem good to me -- we should instead mark permaneant failures with an error (you're right that transient access errors shouldn't be recorded cause the DAGScan itself is retryable) 

Sorry to keep going back and forth on this but I believe there is a class of permaneant errors we want to mark an error for, so that the DAGScan is not retried again just to fail again for the same reason.

Peeja:
> * At the same time, the original DAGScan `DS` will get picked up by IncompleteDAGScansForUpload and run again, and fail again cause the checksum still doesn't match.

Ah, no it won't, because it won't exist anymore. [We delete the DAG scan if it's bad.](https://github.com/storacha/guppy/pull/130/files#diff-7336af42658884d136a0c1cb0daf7f3548d7ed518bb1d5b6bb1de35f79fd14b7R186) That's the point of this PR: we don't need to track the DAG scan state, because we just delete bad ones, because we're not using the old data for anything.

PR #131, opened 2025-09-22
feat: Use better colors for upload progress

Peeja:
Colors are still Storachan, but look a little less error-y.

Also throw in a fancy spinner thing I did over the weekend on a lark because, I mean, why not.


hannahhoward (approved):
LGTM. THIS IS SO COOL!!!

PR #133, opened 2025-09-29
chore: Use Cobra

Peeja:
Also uses Viper, but I haven't actualy tried using it yet. I just let the code generator add it.

#### PR Dependency Tree

* **PR #133** 👈


cmd/largeUpload.go
```diff
@@ -0,0 +1,35 @@
...
+
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+	"github.com/storacha/guppy/cmd/internal/largeupload"
+)
+
+var largeUploadCmd = &cobra.Command{
+	Use:   "large-upload",
```
frrist, line 11: looks like this expects a single argument(?)
`Args:  cobra.ExactArgs(1),` is nice for this.

Edit: I see it's already be used elsewhere, nice!

cmd/largeUpload.go
```diff
@@ -0,0 +1,35 @@
...
+		if err != nil {
+			return err
+		}
+
+		resumeUpload, err := cmd.Flags().GetBool("resume")
+		if err != nil {
+			return err
+		}
+
+		root := cmd.Flags().Arg(0)
```
frrist, line 24: may want to ensure this value isn't empty?

cmd/largeUpload.go
```diff
@@ -0,0 +1,35 @@
...
+
+		return largeupload.Action(cmd.Context(), space, resumeUpload, root)
+	},
+}
+
+func init() {
+	rootCmd.AddCommand(largeUploadCmd)
+
+	largeUploadCmd.Flags().String("space", "", "DID of the space to upload to")
+	largeUploadCmd.Flags().Bool("resume", false, "Resume a previously interrupted upload")
```
frrist, line 34: `StringVar`, `BoolVar`, `<Type>Var` might be nice here for simpler flag parsing.

cmd/largeUpload.go
```diff
@@ -0,0 +1,35 @@
...
+		root := cmd.Flags().Arg(0)
+
+		return largeupload.Action(cmd.Context(), space, resumeUpload, root)
+	},
+}
+
+func init() {
+	rootCmd.AddCommand(largeUploadCmd)
+
+	largeUploadCmd.Flags().String("space", "", "DID of the space to upload to")
```
frrist, line 33: Might want to make this required. iirc there is a cobra command for this.

cmd/root.go
```diff
@@ -0,0 +1,55 @@
...
+		viper.SetConfigFile(cfgFile)
+	} else {
+		// Find home directory.
+		home, err := os.UserHomeDir()
+		cobra.CheckErr(err)
+
+		// Search config in home directory with name ".guppy" (without extension).
+		viper.AddConfigPath(home)
+		viper.SetConfigType("yaml")
+		viper.SetConfigName(".guppy")
```
frrist, line 46: Consider using `os.UserConfigDir()` instead for a more standard config location (we do this over in Piri). e.g. on unix systems this is: `$HOME/.config`, could then have a nest dir called `guppy` containing configuration related stated.

cmd/root.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+func init() {
+	cobra.OnInitialize(initConfig)
+
+	rootCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is $HOME/.guppy.yaml)")
+}
+
+// initConfig reads in config file and ENV variables if set.
+func initConfig() {
```
frrist, line 34: To simplify things I'd consider introducing configuration (files) in a follow on. Using a mixture of vipers different binding methods, which work with cobra: https://github.com/spf13/viper?tab=readme-ov-file#working-with-flags

cmd/up.go
```diff
@@ -0,0 +1,66 @@
...
+	rootCmd.AddCommand(upCmd)
+
+	upCmd.Flags().String("space", "", "DID of the space to upload to")
+	upCmd.Flags().String("proof", "", "Path to archive (CAR) containing UCAN proofs for the operation")
+	upCmd.Flags().StringP("car", "c", "", "Path to CAR file to upload")
+
+	upCmd.Flags().BoolP("hidden", "H", false, "Include paths that start with \".\"")
+	upCmd.Flags().BoolP("json", "j", false, "Format as newline delimited JSON")
+	upCmd.Flags().BoolP("verbose", "v", false, "Output more details")
+	upCmd.Flags().Bool("wrap", true, "Wrap single input file in a directory (Has no effect on directory or CAR uploads. Use --wrap false to disable)")
```
frrist, line 64: nit: this probably existed before this change, but UX preference to avoid flags with a default value of true as it makes their usage confusing imo.

cmd/whoami.go
```diff
@@ -0,0 +1,23 @@
...
+	"fmt"
+
+	"github.com/spf13/cobra"
+	"github.com/storacha/guppy/internal/cmdutil"
+)
+
+var whoamiCmd = &cobra.Command{
+	Use:   "whoami",
+	Short: "Print information about the local agent",
+	Long:  "Prints information about the local agent.",
```
frrist, line 13: might be worth just dropping the `Long` def.

cmd/up.go
frrist: Also might want to make the `--car` and `--wrap` flags mutually exclusive: `cobra.MarkFlagsMutuallyExclusive("car", "warp")` does this I think.

frrist (commented):
left some drive-by comments, good stuff!

Peeja (commented):
@frrist Made changes!

cmd/up.go
```diff
@@ -0,0 +1,42 @@
...
+	rootCmd.AddCommand(upCmd)
+
+	upCmd.Flags().StringVar(&upFlags.spaceDID, "space", "", "DID of the space to upload to")
+	upCmd.Flags().StringVar(&upFlags.proofPath, "proof", "", "Path to archive (CAR) containing UCAN proofs for the operation")
+	upCmd.Flags().StringVarP(&upFlags.carPath, "car", "c", "", "Path to CAR file to upload")
+
+	upCmd.Flags().BoolVarP(&upFlags.hidden, "hidden", "H", false, "Include paths that start with \".\"")
+	upCmd.Flags().BoolVarP(&upFlags.json, "json", "j", false, "Format as newline delimited JSON")
+	upCmd.Flags().BoolVarP(&upFlags.verbose, "verbose", "v", false, "Output more details")
+	upCmd.Flags().BoolVar(&upFlags.wrap, "wrap", true, "Wrap single input file in a directory (Has no effect on directory or CAR uploads. Use --wrap false to disable)")
```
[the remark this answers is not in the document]
Peeja: No no, I mean I'm *literally* removing the entire command in the next PR. 😄 We don't need Guppy to do the JS-client-style upload, and it's currently somewhat broken anyway.

PR #136, opened 2025-10-07
feat: `filecoin/offer` at end of each blob add

Peeja:
#### PR Dependency Tree

* **PR #130**
  * **PR #131**
    * **PR #134**
      * **PR #135**
        * **PR #136** 👈


pkg/preparation/storacha/storacha.go
```diff
@@ -77,10 +95,27 @@ func (a API) SpaceBlobAddShardsForUpload(ctx context.Context, uploadID id.Upload
 			locationCommitment,
 		)
 		if err != nil {
-			return fmt.Errorf("failed to replicate shard %s: %w", shard.CID(), err)
+			return fmt.Errorf("failed to replicate shard %s: %w", shard.ID(), err)
+		}
+
+		pieceDigest, _, err := commpCalc.Digest()
+		if err != nil && !strings.Contains(err.Error(), "insufficient state accumulated") {
```
alanshaw, line 102: Can we just use `shard.Size()` to detect that it's too small and not send it to `commpCalc`?

pkg/preparation/storacha/storacha.go
```diff
@@ -77,10 +95,27 @@ func (a API) SpaceBlobAddShardsForUpload(ctx context.Context, uploadID id.Upload
 			locationCommitment,
 		)
 		if err != nil {
-			return fmt.Errorf("failed to replicate shard %s: %w", shard.CID(), err)
+			return fmt.Errorf("failed to replicate shard %s: %w", shard.ID(), err)
+		}
+
+		pieceDigest, _, err := commpCalc.Digest()
```
alanshaw, line 101: Do we surface this anywhere?

`filecoin/info` uses the piece CID to look up the current deals and inclusion proof.

frrist (commented):
still looking, few comments on thoughts so far

pkg/preparation/storacha/storacha.go
Peeja: We can. I didn't want to depend on the magic number, but this way we're depending on a magic error string, so maybe the magic number and not calling code that won't run is better.

pkg/preparation/storacha/storacha.go
Peeja: Surface as in show the Piece CID to the user? Not currently, in Guppy. It appears in Console, so I assumed we'd eventually make a way to look it up in Guppy as well. Might be worth printing it, though. So far we aren't printing any information about individual shards, except as part of the progress screen.

pkg/preparation/storacha/storacha.go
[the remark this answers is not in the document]
Peeja: Oh, that's exported! Yeah, that's much better. I didn't want to use a magic number, but I don't have to. 🤦🏻‍♀️

Peeja: [Oh, never mind, it's even exported!](https://github.com/storacha/guppy/pull/136#discussion_r2411674601)

pkg/preparation/storacha/storacha.go
```diff
@@ -78,10 +101,32 @@ func (a API) SpaceBlobAddShardsForUpload(ctx context.Context, uploadID id.Upload
...
+			if err != nil {
+				return fmt.Errorf("failed to get piece digest for shard %s: %w", shard.ID(), err)
+			}
+
+			shardPieceCID, err := commcid.DataCommitmentToPieceCidv2(pieceDigest, shard.Size())
+			if err != nil {
+				return fmt.Errorf("failed to get piece CID for shard %s: %w", shard.ID(), err)
+			}
+
+			_, err = a.Client.FilecoinOffer(ctx, spaceDID, cidlink.Link{Cid: shard.CID()}, cidlink.Link{Cid: shardPieceCID})
```
frrist: ❓ do we want to make any assertions on these receipts?

Peeja: I don't think so; that's not a receipt, it's a `filecoincap.OfferOk`. It only contains data we passed in ourselves. If the invocation fails, we'll get an error on `err`.

pkg/preparation/storacha/storacha.go
```diff
@@ -59,12 +66,28 @@ func (a API) SpaceBlobAddShardsForUpload(ctx context.Context, uploadID id.Upload
...
 		if err != nil {
-			return fmt.Errorf("failed to add shard %s to space %s: %w", shard.CID(), spaceDID, err)
+			return fmt.Errorf("failed to add shard %s to space %s: %w", shard.ID(), spaceDID, err)
+		}
+		err = shard.Added(digest)
+		if err != nil {
+			return fmt.Errorf("failed to mark shard %s as added: %w", shard.ID(), err)
 		}
 
 		_, _, err = a.Client.SpaceBlobReplicate(
```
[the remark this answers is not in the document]
Peeja: Same here.

PR #148, opened 2025-11-13
feat: `Locator` can locate blocks

Peeja:
#### PR Dependency Tree

* **PR #148** 👈
  * **PR #149**
    * **PR #150**
      * **PR #152**

### DIFF
--- a/pkg/client/locator/indexlocator.go
+++ b/pkg/client/locator/indexlocator.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"sync"
 
 	mh "github.com/multiformats/go-multihash"
 	"github.com/storacha/go-libstoracha/blobindex"
@@ -29,6 +30,7 @@ func NewIndexLocator(indexer IndexerClient, delegations []delegation.Delegation)
 type indexLocator struct {
 	indexer     IndexerClient
 	delegations []delegation.Delegation
+	mu          sync.RWMutex
 	commitments map[string][]ucan.Capability[assert.LocationCaveats]
 	inclusions  map[string][]inclusion
 }
@@ -73,6 +75,9 @@ func (s *indexLocator) Locate(ctx context.Context, spaceDID did.DID, hash mh.Mul
 }
 
 func (s *indexLocator) getCached(spaceDID did.DID, hash mh.Multihash) ([]Location, bool) {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+
 	inclusions, hasInclusions := s.inclusions[cacheKey(spaceDID, hash)]
 	if !hasInclusions {
 		return nil, false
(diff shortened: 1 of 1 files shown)


hannahhoward (approved):
LGTM with one requested fast follow and several slower follows :P

Question:

I'd like to understand our goals around caching here.
Assuming this is used multiple times during a DAG Walk?

If so, there is a case that is definitely not handled quite optimally:

1. when I query the indexer for a content hash, it will return to me at least:
a. the index for the DAG that contains the content hash
b. the location commitment for the shard within the index that contains the content hash
BUT it will not return the location claims for the other shards

So when I cross from one shard to another, I have the slice information and the shard hash, but no location commitment for the shard.

The proper and optimal way to handle this is to make a location ONLY query to the indexer for the shard that contains the content hash. You can see how the JS version does this here: https://github.com/storacha/blob-fetcher/blob/main/src/locator/index.js#L126 which is a much smaller query cause it doesn't transfer the index again.

However, I think what you have will still work, it's just not optimal. Let's fast follow this one though.

There are various other cases not covered here, which probably isn't critical at the moment but perhaps could use some attention in the future:

Multithreading:
- This definitely is not thread safe, so make sure only one go routine is operating at the same time. Otherwise, add mutexes for all the maps.
- Alternatively, I don't know how long we want this cache to live -- perhaps we just throw it on disk or in the SQL DB.
- If you do want to make it thread safe, you'll also want to track queries in progress to avoid doing additional work

Single block results for legacy claims and compressed queries:
- In prod, we have a variety of legacy ways of indexing things, most of which produce a single location claim just for the hash
- We also have a special compressed query that has the indexer synthesize this kind of claim from an index (useful for minimizing data transfer for HEAD requests)
- We can probably ignore this scenario for now cause Forge will be all new data but will want to add it for hot storage compatibility

Blocks in multiple shards:
- a block can absolutely appear in multiple shards
- looks like the way you deal with this is to track it by hash + space ID, which I think is unique IF and only if it was uploaded via Guppy
- again fine for now, but we should handle this eventually for compatibility.

Peeja:
@hannahhoward 

> The proper and optimal way to handle this is to make a location ONLY query to the indexer for the shard that contains the content hash.

Awesome. I had completely forgotten that was an option. Will fast-follow.

> This definitely is not thread safe, so make sure only one go routine is operating at the same time. Otherwise, add mutexes for all the maps.

Ah, I hadn't caught on that maps are inherently not threadsafe. Updated with a single mutex to cover the (now) two maps, since they're always read and written right next to each other. I could move the locks in closer, but I doubt there's substantial time spent in the intervening space; the queries are outside the lock, and that's what really matters.

> Single block results for legacy claims and compressed queries

Sounds like I'll have to learn more about this for when we do it. I'm intending this command to be Forge-only for this week at least, to keep us moving. Where would you like us to track further work? I'm keeping my own list for now, but if there's a better place I'll put it there.

> Blocks in multiple shards

Fixed! I just had my head twisted around. Thanks for pointing it out. This should make a lot more sense. The locator tracks inclusions it knows about from the indexes and location commitments it knows about for each shard. Then ties the two together to return possibly-multiple locations, which are a commitment (which is a shard-at-a-node) and a position in the shard.

PR #149, opened 2025-11-13
feat: DAG-backed `fs.FS` implementation

Peeja:
#### PR Dependency Tree

* **PR #149** 👈
  * **PR #150**
    * **PR #152**

### DIFF
--- a/pkg/dagfs/dagfs.go
+++ b/pkg/dagfs/dagfs.go
@@ -6,6 +6,7 @@ import (
 	"io/fs"
 	"path"
 
+	"github.com/ipfs/boxo/ipld/merkledag"
 	"github.com/ipfs/boxo/ipld/unixfs"
 	uio "github.com/ipfs/boxo/ipld/unixfs/io"
 	"github.com/ipfs/go-cid"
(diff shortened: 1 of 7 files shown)


hannahhoward (commented):
Broadly, excellent work.

Seperately, and again this is where we get into longer term stuff:

It makes me so sad to see DAGService here instead of just using go-ipld-prime + the UnixFS ADL on top of a block store. 

Fun fact: the main implementation of DAGService is you'll probably use elsewhere is https://github.com/ipfs/boxo/tree/main/ipld/merkledag -- which is built on top of a compatibility shim on top of IPLD prime (github.com/ipfs/go-ipld-legacy) to make it behave like go-ipld-format's notion of a node. EEEK. Worst part is when you look at the author of go-ipld-legacy.

But also, the thing that's really interesting about go-unixfsnode is it allows you to treat a filesystem like an IPLD node, where directories are maps where the keys are filename/dirnames -- this means you can build traversals that work for pure IPLD plus for filesystems.

It is a good time probably to introduce things like Lassie + trustlessutils, and selectors:
https://github.com/ipld/go-trustless-utils -- for implementing an HTTP trustless gateway using go-ipld-prime
https://github.com/filecoin-project/lassie -- a giant retrieval client that services a trustless gateway backed by other retrieval requests -- very close to what we will need to implement eventually... though probably way too complex cause it also implements retrieval over multiple protocols.

PR #154, opened 2025-11-17
fix: Allow blocks to appear more than once

Peeja:
Even simpler than I thought. Literally just don't include the path in the find.

#### PR Dependency Tree

* **PR #154** 👈
  * **PR #155**
    * **PR #156**
      * **PR #167**
        * **PR #170**


hannahhoward:
Also two sources in the same space will have this problem too.

PR #155, opened 2025-11-18
feat: Use Forge's max shard size as default

Peeja:
#### PR Dependency Tree

* **PR #155** 👈
  * **PR #156**
    * **PR #167**
      * **PR #170**

### DIFF
--- a/pkg/preparation/spaces/model/space.go
+++ b/pkg/preparation/spaces/model/space.go
@@ -9,13 +9,13 @@ import (
 )
 
 // MaxShardSize is the maximum allowed size for a shard
-const MaxShardSize = 1 << 29 // 512MB
+const MaxShardSize = 512 << 20 // 512MB
 
 // MinShardSize is the minimum allowed size for a shard
 const MinShardSize = 128
 
 // DefaultShardSize is the default size for a shard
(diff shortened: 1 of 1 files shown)


pkg/preparation/spaces/model/space.go
```diff
@@ -9,13 +9,13 @@ import (
 )
 
 // MaxShardSize is the maximum allowed size for a shard
-const MaxShardSize = 1 << 29 // 512MB
+const MaxShardSize = 512 << 20 // 512MB
```
frrist: The current max shard size is defined here: https://github.com/storacha/piri/blob/main/pkg/pdp/proof/proof.go#L114 this change will cause failures until piri is adjusted. cc @hannahhoward iirc you mentioned something about this being changed in Piri?

alanshaw: That value is the max _with padding_, I believe the value here should be the max unpadded size (266,338,304):

https://gist.github.com/alanshaw/be76c3d4ff555c3a0ee9f5b6e96b5436

Peeja: I guess I got my commits messed up. It should be 254MB. Fixed.

PR #156, opened 2025-11-18
refactor: Warn a bit less

Peeja:
#### PR Dependency Tree

* **PR #156** 👈
  * **PR #167**
    * **PR #170**

### DIFF
--- a/pkg/preparation/uploads/uploads.go
+++ b/pkg/preparation/uploads/uploads.go
@@ -136,6 +136,7 @@ func runScanWorker(ctx context.Context, api API, uploadID id.UploadID, spaceDID 
 		attribute.String("upload.id", uploadID.String()),
 		attribute.String("space.did", spaceDID.String()),
 	))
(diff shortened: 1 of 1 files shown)


pkg/client/spaceblobadd.go
```diff
@@ -220,7 +219,7 @@ func (c *Client) SpaceBlobAdd(ctx context.Context, content io.Reader, space did.
 				return AddedBlob{}, fmt.Errorf("unexpected capability in accept task: %s", ability)
 			}
 		default:
-			log.Warnf("ignoring receipt for unexpected task: %s", concludeRcpt.Ran().Link())
+			log.Infof("ignoring receipt for unexpected task: %s", concludeRcpt.Ran().Link())
```
alanshaw: Hmm, are you seeing this? What is that invocation? Can we warn if we don't know what it is but ignore/not log if we do?

PR #167, opened 2025-11-19
feat: Request blocks/slices individually

Peeja:
Needs some benchmarking. So far it seems to perform significantly worse, taking twice as long. But we need some data for larger volumes of data, as well as memory usage.

#### PR Dependency Tree

* **PR #167** 👈
  * **PR #170**
    * **PR #175**

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,6 +16,7 @@ require (
 	github.com/ipfs/boxo v0.30.0
 	github.com/ipfs/go-block-format v0.2.1
 	github.com/ipfs/go-cid v0.5.0
+	github.com/ipfs/go-datastore v0.8.2
 	github.com/ipfs/go-ipld-cbor v0.2.0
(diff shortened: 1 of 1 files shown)


alanshaw:
> So far it seems to perform significantly worse, taking twice as long.

Just calling out that this is not unexpected. There's no parallel block fetching here and no batching.

It is however a lot less wasteful in egress and memory in the case that you're picking a small file out of a big shard.

The thing that freeway does that allows it to be faster is it makes _multipart_ byte range requests. We can't do that in forge (multipart is not supported/implemented), but we can request multiple blocks in a single range request and split them out from the response. However it's complicated and there's some waste (due to CIDs inbetween block data in the CAR) and it doesn't work well when requesting 2 blocks that have low proximity.

Filepack would help here as blocks will be sequential.

Issue #162, opened 2025-11-19
Adding a source that already exists should not be allowed


frrist:
Steps to reproduce:
```
frrist@cypress ~/w/s/g/s/guppy (frrist/fix/notty)> ./guppy upload sources list did:key:z6MkiXUtThzbbRWUsno5S8U4jvzPu6DwLYoRuA1hvBeTzd2v
Sources for space did:key:z6MkiXUtThzbbRWUsno5S8U4jvzPu6DwLYoRuA1hvBeTzd2v:
- ./upload-test/
- ./2test/

frrist@cypress ~/w/s/g/s/guppy (frrist/fix/notty)> ./guppy upload sources add did:key:z6MkiXUtThzbbRWUsno5S8U4jvzPu6DwLYoRuA1hvBeTzd2v ./2test/

frrist@cypress ~/w/s/g/s/guppy (frrist/fix/notty)> ./guppy upload sources list did:key:z6MkiXUtThzbbRWUsno5S8U4jvzPu6DwLYoRuA1hvBeTzd2v
Sources for space did:key:z6MkiXUtThzbbRWUsno5S8U4jvzPu6DwLYoRuA1hvBeTzd2v:
- ./upload-test/
- ./2test/
- ./2test/ <- wrong?

frrist@cypress ~/w/s/g/s/guppy (frrist/fix/notty)> ./guppy upload  did:key:z6MkiXUtThzbbRWUsno5S8U4jvzPu6DwLYoRuA1hvBeTzd2v
Shards:

Added to Shards:

Scanning Files:



Press q to quit.
2025-11-18T16:08:03.441-0800    FATAL   main    guppy/main.go:22        upload failed: command failed to execute upload: worker encountered an error: bad DAG scan 252bfc78-d323-4a0a-a3a9-43546117dfa5: building UnixFS file: visiting node: creating raw node: constraint failed: UNIQUE constraint failed: nodes.cid, nodes.space_did (1555)
```

Pranav-1100:
Hi @frrist, Can I work on this issue, I verified this issue locally, I found the issue in uploadSourcesAddCmd in cmd/upload.go.

frrist:
This is already being worked on here: https://github.com/storacha/guppy/pull/171

PR #171, opened 2025-11-20
fix: sources  cannot contain duplicate paths

frrist:
- e.g. can't add the same path as a source twice.
- When creating a source that already exists via CreateSource the existing source will be returned.

Implementing this change has caused me to question of constraints of the source table. Currently it uses a uuid as the primary key, perhaps a the `path` and `kind` field should instead be a composite primary key?

closes https://github.com/storacha/guppy/issues/162

### DIFF
--- a/pkg/preparation/sqlrepo/sources.go
+++ b/pkg/preparation/sqlrepo/sources.go
@@ -25,8 +25,8 @@ func (r *Repo) CreateSource(ctx context.Context, name string, path string, optio
 	}
 
 	// reuse an existing source pointing at the same path if one already exists
-	// NB(forrest): one intention of the source abstraction is that they mau have different kinds
-	// at present, there is only a local time, so the existence check on a normalized path here is sufficient.
+	// NB(forrest): one intention of the source abstraction is that they may have different kinds
+	// at present, there is only a local kind, so the existence check on a normalized path here is sufficient.
 	// In the future when sources may have different kinds this check will need adjustment.
 	existing, err := r.getSourceByPath(ctx, absPath)
 	if err != nil {


pkg/preparation/sqlrepo/sources.go
```diff
@@ -17,7 +18,25 @@ var _ sources.Repo = (*Repo)(nil)
...
 func (r *Repo) CreateSource(ctx context.Context, name string, path string, options ...sourcemodel.SourceOption) (*sourcemodel.Source, error) {
-	src, err := sourcemodel.NewSource(name, path, options...)
+	// resolve the specific path to its absolute path
+	absPath, err := filepath.Abs(path)
+	if err != nil {
+		return nil, fmt.Errorf("failed to resolve absolute path from %s: %w", path, err)
+	}
+
+	// reuse an existing source pointing at the same path if one already exists
+	// NB(forrest): one intention of the source abstraction is that they mau have different kinds
```
alanshaw:
```suggestion
	// NB(forrest): one intention of the source abstraction is that they may have different kinds
```

frrist:
> What happens if you specify a path within a source? I assume that would currently work? Shouldn't we also deny that...or return the source it is in?

Right now, nothing the source will be added. I'm not sure what the right answer is here specifically, but I think the current behavior seems right.

But at a higher level, I suspect we'll need some check on the sources being added to a space. Since we probably want to avoid something like:
```
Space:bafywhatever
 - Sources:
   - /A/
   - /A/B/
```
i.e. we shouldn't allow adding sources to a space whos parent is already in the space

PR #172, opened 2025-11-20
feat: implement configuration system

frrist:
- refactor command layout


pkg/repo/space_store.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+
+	// Grant all necessary capabilities for the space
+	capabilities := []ucan.Capability[ucan.NoCaveats]{
+		ucan.NewCapability("space/*", s.DID.String(), ucan.NoCaveats{}),
+		ucan.NewCapability("upload/*", s.DID.String(), ucan.NoCaveats{}),
+		ucan.NewCapability("store/*", s.DID.String(), ucan.NoCaveats{}),
+		ucan.NewCapability("index/*", s.DID.String(), ucan.NoCaveats{}),
+		ucan.NewCapability("filecoin/*", s.DID.String(), ucan.NoCaveats{}),
+		ucan.NewCapability("blob/*", s.DID.String(), ucan.NoCaveats{}),
```
alanshaw: Hmm, I thought we delegated `*` to the email?

If not, can we please add `assert/*` here like https://github.com/storacha/guppy/pull/141

frrist: Humm seems like a bug then. fwiw this code was moved from https://github.com/storacha/guppy/blob/main/cmd/space.go#L439

so sounds like this should instead mimic/reference the capabilities as they are defined in your link: https://github.com/storacha/guppy/blob/a3a82d77e28ef1892f5ab9864a176764c19d040f/pkg/client/requestaccess.go#L13 ?

pkg/repo/space_store.go
```diff
@@ -0,0 +1,232 @@
...
+	_, ok := state.Spaces[name]
+	if ok {
+		return Space{}, fmt.Errorf("space '%s' already exists", name)
+	}
+	sk, err := signer.Generate()
+	if err != nil {
+		return Space{}, err
+	}
+
+	privKeyStr, err := multibase.Encode(multibase.Base64pad, sk.Encode())
```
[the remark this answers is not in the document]
frrist: also a move from here: https://github.com/storacha/guppy/blob/main/cmd/space.go#L324 (but glad we are catching these things) will adjust.

Reason for the move: Trying to get everything that depends on a file system path to be handled via the repo package, rather than calls like this: https://github.com/storacha/guppy/blob/main/cmd/space.go#L337 scattered throughout the codebase

pkg/repo/space_store.go
alanshaw: @Peeja just checking if this was intentional?

frrist:
Too much to rebase here; nearly everything is different since this was opened, closing.

PR #175, opened 2025-11-25
feat: Space management, redux

Peeja:
#### PR Dependency Tree

* **PR #175** 👈
  * **PR #180**
  * **PR #187**
    * **PR #198**

### DIFF
--- a/cmd/space/generate.go
+++ b/cmd/space/generate.go
@@ -61,19 +61,23 @@ var generateCmd = &cobra.Command{
 		if generateFlags.provisionTo != "" {
 			provisionAccount, err = didmailto.FromInput(generateFlags.provisionTo)
 			if err != nil {
+				cmd.SilenceUsage = false
 				return fmt.Errorf("parsing `--provision-to` account %q: %w", generateFlags.provisionTo, err)
 			}
 			if !slices.Contains(accounts, provisionAccount) {
-				return fmt.Errorf("account %s is not logged in yet. Use `guppy login %s` to log in.", generateFlags.provisionTo, generateFlags.provisionTo)
+				fmt.Printf("Account %s is not logged in yet. Use `guppy login %s` to log in.", generateFlags.provisionTo, generateFlags.provisionTo)
+				return cmdutil.NewHandledCliError(fmt.Errorf("account %s is not logged in", provisionAccount))
 			}
 		} else {
 			switch {
 			case len(accounts) == 0:
-				return fmt.Errorf("No accounts are logged in yet. Use `guppy login <account>` to log in.")
+				fmt.Printf("No accounts are logged in yet. Use `guppy login <account>` to log in.")
+				return cmdutil.NewHandledCliError(fmt.Errorf("account %s is not logged in", provisionAccount))
 			case len(accounts) == 1:
 				provisionAccount = accounts[0]
 			default:
-				return fmt.Errorf("Multiple accounts are logged in. Specify an account with `--provision-to`.")
+				fmt.Printf("Multiple accounts are logged in. Specify an account with `--provision-to`.")
+				return cmdutil.NewHandledCliError(fmt.Errorf("multiple accounts are logged in"))
 			}
 		}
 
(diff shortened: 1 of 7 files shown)


cmd/space/generate.go
```diff
@@ -0,0 +1,65 @@
...
+var generateCmd = &cobra.Command{
+	Use:   "generate",
+	Short: "Generate a new space",
+	Long: wordwrap.WrapString(
+		"Generates a new Storacha space, and stores it in the local store.",
+		80),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		space, err := signer.Generate()
+		if err != nil {
+			cmd.SilenceUsage = false
```
alanshaw: Why is this?

cmd/space/grant.go
```diff
@@ -0,0 +1,135 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("storing delegation: %w", err)
+	}
+
+	return delToStore, nil
+}
+
+func init() {
+	grantCmd.Flags().StringSliceVar(&grantFlags.capabilities, "can", spaceAccess,
+		"Capabilities to grant (e.g., 'store/add', 'upload/*'). If not specified, grants default space access.")
```
alanshaw: I was under the impression that we granted admin `*` to the account?...and we grant "space access" from the account to the agent.

cmd/space/generate.go
```diff
@@ -0,0 +1,65 @@
...
+			return fmt.Errorf("generating signer for space: %w", err)
+		}
+
+		c := cmdutil.MustGetClient(*StorePathP)
+
+		err = c.AddSpace(space)
```
alanshaw: Should we check we have an account to register it with before we store the space in the client? I would basically do all of create, provision and delegation before adding it to the local store. Otherwise you end up with an unusable space...

cmd/account/ls.go
```diff
@@ -0,0 +1,43 @@
...
+
+		// Build JSON array of account DIDs
+		output := make([]map[string]string, 0, len(accounts))
+		for _, account := range accounts {
+			output = append(output, map[string]string{
+				"id": account.String(),
+			})
+		}
+
+		jsonBytes, err := json.Marshal(output)
```
alanshaw: Maybe use `MarshalIndent` if this is the default output...consider just outputting the DIDs and adding a flag for JSON output?

cmd/space/generate.go
```diff
@@ -0,0 +1,65 @@
...
+		_, err = grant(cmd.Context(), c, space, account)
+		if err != nil {
+			return fmt.Errorf("granting capabilities: %w", err)
+		}
+
+		// Print JSON output with just the DID (omitting the private key)
+		output := map[string]string{
+			"id": space.DID().String(),
+		}
+		jsonBytes, err := json.Marshal(output)
```
alanshaw: `MarshalIndent` or more human friendly output?

pkg/client/options.go
```diff
@@ -54,3 +55,13 @@ func WithSaveFn(saveFn func(agentdata.AgentData) error) Option {
 		return nil
 	}
 }
+
+// WithAdditionalProofs adds proofs to the client that will be included in
+// Proofs() results but will not be saved to the client's store. This is
+// useful for proofs that are only needed for a single operation.
```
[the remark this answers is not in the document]
Peeja: Isn't that basically the same thing, or a case of it? Also, for a truly ephemeral client, you can do this, _or_ you can use `.AddProofs()`; it's not going to be saved to disk. `.AddProofs()` just means you'll keep the proofs for additional operations with that client, in that process.

cmd/space/generate.go
```diff
@@ -0,0 +1,181 @@
...
+		spaceSigner,
+		c.Issuer().DID(),
+		capabilities,
+		delegation.WithNoExpiration(),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("creating delegation: %w", err)
+	}
+
+	c.AddProofs(delToStore, delToKeep)
```
frrist: missing an error check

cmd/ls.go
```diff
@@ -30,6 +31,7 @@ var lsCmd = &cobra.Command{
 	RunE: func(cmd *cobra.Command, args []string) error {
 		spaceDID, err := did.Parse(cmd.Flags().Arg(0))
 		if err != nil {
+			cmd.SilenceUsage = false
```
frrist: Here, and many other places: Why do this? The default value of an unset bool is false. Unless we have overridden it somewhere else?

hannahhoward: +1 not sure why it's needed to set this over and over

cmd/root.go
```diff
@@ -36,13 +42,13 @@ func init() {
...
 		panic(fmt.Errorf("failed to get user home directory: %w", err))
 	}
-	defaultStorachaDir := filepath.Join(homedir, ".storacha")
-
-	uploadCmd.PersistentFlags().StringVar(&uploadDbPath, "storacha-dir", defaultStorachaDir, "Directory containing the config and data store (default: ~/.storacha)")
 
-	// We could make this configurable in the future, but for now it seems like
-	// too many options for likely no value.
-	storePath = filepath.Join(defaultStorachaDir, "store.json")
+	rootCmd.PersistentFlags().StringVar(
```
hannahhoward: gladd to see this one appear :)

cmd/space/generate.go
```diff
@@ -0,0 +1,181 @@
...
+
+	// Store the delegation via access/delegate
+	_, err = c.AccessDelegate(ctx, spaceSigner.DID(), delToStore)
+	if err != nil {
+		return nil, fmt.Errorf("storing delegation: %w", err)
+	}
+
```
hannahhoward: Nit for hot storage we need to grant access to the gateway to serve... tomorrow problem though

hannahhoward:
also this PR is great. so nice to see Guppy get full featured

PR #210, opened 2025-12-02
feat(preparation): improve retriability

hannahhoward:
# Goals

I discovered by accident that if I stopped a 2GB simulated upload half way, then replaced all the data in the directory with a different 2GB structure (using randdir)... well, it was extremely hard, in fact impossible to resume all the way to completion. There were essentially two problems:
- Each run only potentially resolves one error, but there are potentially thousands of errors if the whole structure changes
- Some errors that should be recoverable became impossible to recover from -- such as shards that got too small to upload at all (or ran out of CIDs entirely)

This PR is the result of all my attempts to make this work smoothly, and now it does.

# Implementation

- adds an auto retry mode -- if you pass `--retry` to upload, it will automatically restart the upload until success
- properly marks bad FS entries as retriable
- identifies retriable errors (bad FS Entries, bad Nodes) by wrapping them in a RetriableErr struct
- makes more progress building dags before a failure - now a full iteration of dags scans run before returning a potentially multiple bad fs entries
- properly handles bad shards by not only removing items but putting the good CIDs back in the queue and deleting the shard. This is important cause the shard can get too small and potentially break if everything is emptied out.

# For Discussion

- THere's a schema.sql change in there. This is probably the last go we'll get without migrations
- should it be a number of retries?

### DIFF
--- a/cmd/internal/upload/demo/demo.go
+++ b/cmd/internal/upload/demo/demo.go
@@ -356,5 +356,5 @@ func Demo(ctx context.Context, repo *sqlrepo.Repo, spaceName string, alterMetada
 
 	}
 
-	return ui.RunUploadUI(ctx, repo, api, uploads)
+	return ui.RunUploadUI(ctx, repo, api, uploads, false)
 }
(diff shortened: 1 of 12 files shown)


frrist (approved):

Peeja:
> * properly handles bad shards by not only removing items but putting the good CIDs back in the queue and deleting the shard. This is important cause the shard can get too small and potentially break if everything is emptied out.

Ooooh, fantastic catch.

PR #229, opened 2025-12-03
fix: Don't add links for directory children from other spaces

Peeja:
This was nasty to track down.

If you reuse the same DB and upload the same directory multiple times, you'll end up adding the directory children from _previous_ runs to your current UnixFS directory. Of course, UnixFS is fine with this; it's not until you try to _retrieve_ the data and write it to disk that you have a problem, because you're trying to write the same file twice. UnixFS directories can contain the same file multiple times, but not real ones.

The solution is to add a missing space check. As soon as we descend into a directory and look up its children, we're picking up the children from previous runs, and generating links from those. This fixes that.

### DIFF
--- a/pkg/preparation/sqlrepo/dags.go
+++ b/pkg/preparation/sqlrepo/dags.go
@@ -190,7 +190,7 @@ func (r *Repo) DirectoryLinks(ctx context.Context, dirScan *model.DirectoryDAGSc
 		FROM directory_children
 		JOIN fs_entries ON directory_children.child_id = fs_entries.id
 		JOIN dag_scans ON directory_children.child_id = dag_scans.fs_entry_id
-		JOIN nodes ON dag_scans.cid = nodes.cid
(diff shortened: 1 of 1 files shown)


pkg/preparation/sqlrepo/dags.go
```diff
@@ -191,7 +191,8 @@ func (r *Repo) DirectoryLinks(ctx context.Context, dirScan *model.DirectoryDAGSc
 		JOIN fs_entries ON directory_children.child_id = fs_entries.id
 		JOIN dag_scans ON directory_children.child_id = dag_scans.fs_entry_id
 		JOIN nodes ON dag_scans.cid = nodes.cid
-		WHERE directory_children.directory_id = ?`
+		WHERE directory_children.directory_id = ?
+		AND nodes.space_did = fs_entries.space_did`
```
hannahhoward: shouldn't this be part of the JOIN?

Issue #226, opened 2025-12-03
Guppy should use latest indexing client


Peeja:
This will handle error messages from the indexer correctly (decoding gzip).

Depends on https://github.com/storacha/go-libstoracha/pull/76 landing.

PR #233, opened 2025-12-05
feat: add guppy explore command 

### LINKED ISSUE #222: Guppy should be able to explore a tree from a CID
(Or do we leave that for just the web interface?)

Goddhi:
implements the guppy explore <root-cid> command, allowing users to interactively traverse and list the contents of a UnixFS DAG stored on the network.
i added consumer/get capability because  while implementing the explore feature, I encountered issues where the Indexer (indexer.storacha.network) would return "no locations found" for known valid CIDs.

Upon investigation (and referencing issue #226 ), it became clear that the Indexer now enforces an authorization check requiring the consumer/get capability. it seems the previous space/content/retrieve capability is no longer sufficient for these indexer queries.

To try to resolve this, I added the pkg/capabilities/consumer package to model this new capability.
Usage:
```
guppy explore <root-cid> --space 
```

Closes #222

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -10,9 +10,8 @@ import (
 
 	logging "github.com/ipfs/go-log/v2"
 	"github.com/spf13/cobra"
-	"github.com/spf13/pflag"
-	"github.com/storacha/guppy/cmd/unixfs"
 	"github.com/spf13/viper"
+	"github.com/storacha/guppy/cmd/unixfs"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/trace"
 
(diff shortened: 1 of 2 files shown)


Peeja:
Ah: the capabilities are in https://github.com/storacha/go-libstoracha/pull/76, which I only just merged.

This is good, but might have some issues at larger data sizes. I wasn't really intending anyone to pick up that issue yet; it still needs some thought and has some context behind it that hasn't been written out yet. In general, safer to check in on issues before you start working on them; I wouldn't want you to spend a bunch of effort on something we end up having to throw out because of something you weren't aware of.

PR #246, opened 2025-12-09
fix: retrieval when ~~princess~~ block is in a different castle

alanshaw:
When we have an index and try to fetch a block which is in a shard that we don't have a location claim for, we need to query the indexer for the shard hash NOT the block hash. Querying for the block hash will not give us anything back when query type is "location".


hannahhoward (approved):
LGTM

PR #253, opened 2025-12-10
allow output of `space generate` to re-direct to env var

frrist:
- space goes to stdout, all other info to sdterr
- enables this sort of thing: `export SPACE=(./guppy space generate)`


hannahhoward (approved):
LGTM pending fixing checks

PR #255, opened 2025-12-10
feat: authorize retrieval on demand

alanshaw:
This PR authorizes index retrieval on demand, including only the proofs needed for the invocation, not ALL proofs.

It also allows the indexer client to be used in longer running processes that may outlive the expiration of delegations created when the indexer client is constructed.

Finally, requests for content in multiple spaces can be autorized as required without redundant delegations.


hannahhoward (approved):
shipit

PR #260, opened 2025-12-12
refactor(storacha): seperate blob post process into seperate worker

hannahhoward:
# Goals

Speed up uploads by handling post processing steps for shards (replication/filecoin offer) after uploading in a separate step from running uploads.

Tests show that even at high parallelism bandwidth used can become inconsistent because so much time is spent waiting for the post process steps (specifically filecoin/offer) to complete. Here we seperate into a seperate bucket so as soon as upload completes, another upload can begin.

# Implementation

- Make "uploaded" a seperate state (arguably this could be computed from the presence/absence of a location claim, but that makes the query quite a bit more complicated)
- Seperate AddShardsForUpload into two steps: AddShardsForUpload and PostProcessUploadedShards
- Add a seperate working for post processing

### DIFF
--- a/pkg/preparation/blobs/model/blob.go
+++ b/pkg/preparation/blobs/model/blob.go
@@ -21,6 +21,8 @@ const (
 	// BlobStateClosed indicates that a blob is no longer accepting nodes, but
 	// is not yet added to the space.
 	BlobStateClosed BlobState = "closed"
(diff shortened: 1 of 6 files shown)


hannahhoward (commented):
LGTMing based on prod results

PR #261, opened 2025-12-12
fix(sqlrepo): add index for shards

hannahhoward:
add an index for shards by upload_id and state to make ShardsForUploadByState faster

also includes a fix for our flaky test that is failing CI on several branches


pkg/preparation/sqlrepo/migrations/20251212174118_add_upload_id_state_index_to_shard.sql
frrist: We may also want an index on the nodes_in_shards table:
```sql
CREATE INDEX IF NOT EXISTS idx_nodes_in_shards_shard_id_offset
  ON nodes_in_shards (shard_id, shard_offset);
```
so `NodesByShard`/`ForEachNode` can satisfy `WHERE shard_id = ? ORDER BY shard_offset` via the index instead of scanning.

PR #262, opened 2025-12-12
feat: Indexes upload separately, as available

Peeja:
One pretty unfortunate thing here: we can't *actually* add the indexes as the shards become available for indexing (ie, closed), we have to wait until we have a root CID, which means _all_ the shards are closed. That's because the root CID is listed in the index, and in the `space/index/add` invocation. That said, I'm pretty sure it doesn't need to be, and we'll respect the index no matter what that's set to. After all, you can request a subdag just fine, without knowing the root CID of the larger DAG it was uploaded as a part of.

If we relax this restriction in the future, we can make the index uploads start sooner with a small change to the pipeline code: we just need to signal as shards close instead of at the end of the DAG scan.

But this PR _does_ still mean that the indexes aren't held in memory all at once, and that they can be added in parallel with the shards being added (just not in parallel with shards being *assembled*—they have to all be closed, but don't have to be added yet).

#### PR Dependency Tree

* **PR #262** 👈


hannahhoward:
Parallelism issue, fixed in main, adjusted here

hannahhoward:
@Peeja I did a sizable refactor because indexes are shards are so similar -- basically I extracted all the common functionality and made the shards module a blobs module. It feels a lot cleaner

Otherwise, everything was great, so I'm merging.

hannahhoward (approved):
LGTM

Peeja:
Oh, that's brilliant! I'm so influenced by ActiveRecord from…uh, decades ago (*shudder*) that I didn't even think to use one model type with multiple tables. 😅

PR #266, opened 2025-12-17
feat: support http3/QUIC with fallback in client

frrist:


frrist:
closing as this didn't yeild a performance improvment in practice, it actually made things slower.

PR #268, opened 2025-12-19
feat: Add otel counters to the stages

Peeja:
#### PR Dependency Tree

* **PR #268** 👈


hannahhoward (changes requested):
a couple renames I'd like to maintain consistency.

also, do we need to make the guppy client did an attribute on each of these -- otherwise I think we'll just get a mega counter for all guppy instances?

PR #270, opened 2025-12-23
fix(sqlrepo): fix various bugs with DB ops

hannahhoward:
# Goals

We're seeing periodic big increases in WAL file size and SQLITE_BUSY messages

# Implementation

- Despite many attempts with google searching and multiple AI models, it's really not 100% clear why the WAL size is growing, other than a whole lot of things are happening at once. There are clear reasons a WAL size COULD grow rapidly but most have to do with concurrency -- which is hard to figure out given our call to SetMaxOpenConns(1). So in the absense of that, just forcing a WAL size compaction periodically seems a reasonable compromise. (there is also the possibility to leave a query open by not calling rows.Close() but I've audited the code for this issue.
- The SQLITE_BUSY call is also confusing, though I suspect also related to a very large WAL size. In any case 3 seconds is on the small side for this type of database, so I think perhaps better to increase it.
- The last call is an operation to transactionalize and simplify the add shard to index call.

#### PR Dependency Tree

* **PR #270** 👈
  * **PR #271**
    * **PR #272**
      * **PR #273**
        * **PR #274**
          * **PR #275**
            * **PR #276**
              * **PR #277**

### DIFF
--- a/pkg/preparation/uploads/uploads.go
+++ b/pkg/preparation/uploads/uploads.go
@@ -104,6 +104,21 @@ func signal(work chan<- struct{}) {
 	}
 }
 
+func multiplexSignal(input <-chan struct{}, outputs ...chan<- struct{}) {
+	// Multiplex closed shards to both indexing and uploading workers
+	go func() {
+		for range input {
+			for _, out := range outputs {
+				signal(out)
+			}
+		}
+
+		for _, out := range outputs {
+			close(out)
+		}
+	}()
+}
+
 // ExecuteUpload executes the upload process for a given upload, handling its state transitions and processing steps.
 func (a API) ExecuteUpload(ctx context.Context, uploadID id.UploadID, spaceDID did.DID) (cid.Cid, error) {
 	ctx, span := tracer.Start(ctx, "execute-upload", trace.WithAttributes(
@@ -128,15 +143,7 @@ func (a API) ExecuteUpload(ctx context.Context, uploadID id.UploadID, spaceDID d
 	)
 
 	// Multiplex closed shards to both indexing and uploading workers
-	go func() {
-		for range closedShardsAvailable {
-			signal(shardsNeedIndexing)
-			signal(shardsNeedUploading)
-		}
-
-		close(shardsNeedIndexing)
-		close(shardsNeedUploading)
-	}()
+	multiplexSignal(closedShardsAvailable, shardsNeedIndexing, shardsNeedUploading)
 
 	// Start the workers
 	eg, wCtx := bettererrgroup.WithContext(ctx)


pkg/preparation/blobs/blobs.go
```diff
@@ -419,11 +419,7 @@ func (a API) makeErrBadNodes(ctx context.Context, shardID id.ShardID, nodeReader
 	// Collect the nodes first, because we can't read data for each node while
 	// holding the lock on the database that ForEachNode has. This means holding a
 	// bunch of nodes in memory, but it's limited to the size of a shard.
-	var nodes []dagsmodel.Node
-	err := a.Repo.ForEachNode(ctx, shardID, func(node dagsmodel.Node, _ uint64) error {
-		nodes = append(nodes, node)
-		return nil
-	})
+	nodes, err := a.Repo.NodesByShard(ctx, shardID, 0)
```
Peeja: **praise:** Nice.

pkg/preparation/sqlrepo/dags.go
```diff
@@ -334,7 +334,7 @@ func (r *Repo) HasIncompleteChildren(ctx context.Context, directoryScans *model.
 			SELECT 1
 			FROM dag_scans
 			JOIN directory_children ON directory_children.child_id = dag_scans.fs_entry_id
-			WHERE directory_children.directory_id = X'0874d43f68d443adbfc176020eab51cf'
+			WHERE directory_children.directory_id = ?
```
Peeja: Image

**praise:** Uhhh… good catch.

**question:** How was this not causing problems?

pkg/preparation/sqlrepo/sqlrepo.go
```diff
@@ -29,15 +31,65 @@ func Null[T any](v *T) sql.Null[T] {
...
+	r.StopAutoCheckpoint()
 	return r.db.Close()
 }
+
+// Checkpoint forces a WAL checkpoint to transfer data from the write-ahead log
+// to the main database file. This should be called periodically during long
+// operations to prevent unbounded WAL growth. The TRUNCATE mode resets the WAL
+// file to zero bytes after a successful checkpoint.
+func (r *Repo) Checkpoint(ctx context.Context) error {
+	_, err := r.db.ExecContext(ctx, "PRAGMA wal_checkpoint(TRUNCATE)")
```
Peeja: **suggestion:** Would `RESTART` be more appropriate? There's a cost to actually truncating the file, but SQLite can also just rewind and overwrite the existing WAL to avoid that.

pkg/preparation/sqlrepo/sqlrepo.go
```diff
@@ -29,15 +31,65 @@ func Null[T any](v *T) sql.Null[T] {
...
 type Repo struct {
-	db *sql.DB
+	db             *sql.DB
+	checkpointStop chan struct{}
+}
+
+// StartAutoCheckpoint starts a background goroutine that periodically
+// checkpoints the WAL to prevent unbounded growth during long operations.
+// Call StopAutoCheckpoint to stop it, or it will be stopped when Close is called.
+func (r *Repo) StartAutoCheckpoint(ctx context.Context, interval time.Duration) {
```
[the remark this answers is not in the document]
Peeja: Ah, I see, this is called from `OpenRepo()`. That's starting to make me think that this whole checkpointing thing may be `OpenRepo()`'s job in the first place, though.

pkg/preparation/repo.go
```diff
@@ -16,10 +16,11 @@ import (
 )
 
 const (
-	defaultJournalMode = "WAL"
-	defaultSynchronous = "NORMAL"
-	defaultBusyTimeout = 3 * time.Second
-	defaultForeignKeys = true
+	defaultJournalMode      = "WAL"
+	defaultSynchronous      = "NORMAL"
+	defaultBusyTimeout      = 60 * time.Second
```
Peeja: **question (non-blocking):** What inspired this? It's probably fine, but it mildly concerns me that it was necessary.

Peeja (dismissed):
Looks great in general, just some proposed adjustment.

PR #271, opened 2026-01-01
test(preparation): add a benchmark

hannahhoward:
# Goals

I wanted to try various things to spead up SQL processing (most of which didn't end up helping) so I built this benchmark, which I imagine can still be useful as we try different things.

#### PR Dependency Tree

* **PR #270**
  * **PR #271** 👈
    * **PR #272**
      * **PR #273**
        * **PR #274**
          * **PR #275**
            * **PR #276**
              * **PR #277**

### DIFF
--- a/pkg/preparation/uploads/uploads.go
+++ b/pkg/preparation/uploads/uploads.go
@@ -104,6 +104,21 @@ func signal(work chan<- struct{}) {
 	}
 }
 
+func multiplexSignal(input <-chan struct{}, outputs ...chan<- struct{}) {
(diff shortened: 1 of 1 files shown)


pkg/preparation/preparation_test.go
```diff
@@ -187,19 +189,22 @@ func prepareTestClient(
 }
 
 func createUpload(
-	t *testing.T,
+	t testing.TB,
+	ctx context.Context,
```
Peeja: **suggestion:** Can we keep using `t.Context()`? It looks like that's always what we're passing in anyway.

PR #275, opened 2026-01-01
refactor(blobs): seperate add shards to indexes worker

hannahhoward:
# Goal
seperate process of adding shards to indexes to its own worker

# Implementation

This actually turns out to be really easy via the implementation of a NOT EXISTS query, to find shards that are not in indexes

#### PR Dependency Tree

* **PR #270**
  * **PR #271**
    * **PR #272**
      * **PR #273**
        * **PR #274**
          * **PR #275** 👈
            * **PR #276**
              * **PR #277**

This tree was auto-generated by [Charcoal](https://github.com/danerwilliams/charcoal)

PR #277, opened 2026-01-06
refactor(uploads): minor suggested generalization

hannahhoward:
just way we could generalize this if we need to use it again.

#### PR Dependency Tree

* **PR #270**
  * **PR #271**
    * **PR #272**
      * **PR #273**
        * **PR #274**
          * **PR #275**
            * **PR #276**
              * **PR #277** 👈

### DIFF
--- a/pkg/preparation/uploads/uploads.go
+++ b/pkg/preparation/uploads/uploads.go
@@ -104,6 +104,21 @@ func signal(work chan<- struct{}) {
 	}
 }
 
+func multiplexSignal(input <-chan struct{}, outputs ...chan<- struct{}) {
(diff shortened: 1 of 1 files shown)


pkg/preparation/uploads/uploads.go
```diff
@@ -104,6 +104,21 @@ func signal(work chan<- struct{}) {
 	}
 }
 
+func multiplexSignal(input <-chan struct{}, outputs ...chan<- struct{}) {
+	// Multiplex closed shards to both indexing and uploading workers
```
Peeja: **suggestion:** 😄 

```suggestion
```

PR #278, opened 2026-01-07
feat(preparation): fix preparation test flakyness, which required several fixes

hannahhoward:
The error related to the uploads table in the preparation tests that came up in one of the PRs is something I've seen intermittently on my machine. So I decided to fix it seperately.

Fundamentally, the problem is SQLite in memory databases can get wiped even in shared cached mode if the database connections drop to zero. Ultimately I got tired of trying to debug how that could happen and just switched to using a file based DB.

But that revealed several more race issues that almost surely have occurred in production, which occur intermittently running the preparation restart test. Essentially, a context cancellation can cause a number of issues -- a directory getting created without its links, or a node created without an associated node upload record -- all cause context can get cancelled between node creation and directory link creation or between node creation and node upload creation.

The only real solution is to throw it all in a transaction, which conveniently simplifies a number of things other than the transaction itself.

I think we will probably have to do this for CreateDagScan as well -- put it in a transaction with FindOrCreateFSEntry

Also found another aspect of the test that could be flaky -- the number of shards uploaded after the blob put fails.

And a data race, which I fixed.

### DIFF
--- a/pkg/preparation/blobs/blobs_test.go
+++ b/pkg/preparation/blobs/blobs_test.go
@@ -49,9 +49,7 @@ func TestAddNodeToUploadShardsAndCloseUploadShards(t *testing.T) {
 		openShards, err := repo.ShardsForUploadByState(t.Context(), upload.ID(), model.BlobStateOpen)
 		require.NoError(t, err)
 		require.Len(t, openShards, 0)
-		n, _, err := repo.FindOrCreateRawNode(t.Context(), nodeCID1.(cidlink.Link).Cid, 1<<14, spaceDID, "some/path", source.ID(), 0)
-		require.NoError(t, err)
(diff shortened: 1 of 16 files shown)


hannahhoward:
Yea I agree and... oy I don't know quite what to do. Domain Driven Design and SQL transactions seem almost frustratingly in opposition to each other. 

Pretty sure the proper DDD approach would be to keep node_uploads and nodes_in_shards as seperate tables, which I was looking at one point.

Another move would be to store the upload_id on the node itself, but then it would "initial upload_id" to indicate that it's simply that uploads job to put things in the database. That would enable us to revert to put back node_uploads to essentially how nodes_in_shards was, and to mark.

We should schedule a design discussion to talk about this. There are a few other design issues bugging me.

PR #283, opened 2026-01-08
feat: add delegation create command

alanshaw:
Similar to the JS client:

e.g.

`guppy delegation create <space> <audience> --can upload/add`


Peeja (approved):
Awesome!

PR #284, opened 2026-01-08
feat: add proof add command

alanshaw:
e.g.

```
guppy proof add <data-or-path>
```

...where `<data-or-path>` can be the string printed to the console or a path to the CAR file saved to disk from the [`guppy delegation create ...`](https://github.com/storacha/guppy/pull/283) command.

Adding a space:

Screenshot 2026-01-08 at 16 35 19

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -93,7 +93,7 @@ func (c *Client) Issuer() principal.Signer {
 
 // CapabilityQuery represents a query to filter proofs by capability.
 type CapabilityQuery struct {
-	// Can is the ability to match (e.g., "store/add"). Use "*" to match all abilities.
+	// Can is the ability to match (e.g., "store/add").
 	Can ucan.Ability
 	// With is the resource to match. Omit or set to empty string to query for any
 	// resource. Note: using "ucan:*" will only match "ucan:*" resources.


pkg/client/spaces.go
```diff
@@ -12,11 +10,11 @@ import (
 
 // Spaces returns all spaces we can act as.
 func (c *Client) Spaces() ([]did.DID, error) {
-	return spacesFromDelegations(c.Proofs(CapabilityQuery{Can: "space/*"}))
+	return spacesFromDelegations(c.Proofs(CapabilityQuery{Can: "space/*", With: "ucan:*"}))
```
alanshaw: Another bug - we need to specify _any_ resource to get all spaces for the agent.

pkg/client/client.go
```diff
@@ -233,8 +234,9 @@ func matchesAbility(capAbility, queryAbility ucan.Ability) bool {
 // A capability matches if:
 //   - The resources match exactly
 //   - The capability has "ucan:*" (matches any query resource)
+//   - The query resource is empty (matches any capability resource)
 func matchesResource(capResource, queryResource ucan.Resource) bool {
-	return capResource == "ucan:*" || queryResource == capResource
+	return queryResource == "" || capResource == "ucan:*" || queryResource == capResource
```
alanshaw: This was a bug. If you don't specify a resource in the query it should match any resource.

Previosuly this worked because the only delegation we have is from login, where the delegation resource is `ucan:*`. However if you obtain a delegation for a specific resource `did:key:...` then empty string does not match it.

PR #290, opened 2026-01-15
feat: Tie up less memory with coalesced batches

Peeja:
#### PR Dependency Tree

* **PR #290** 👈
  * **PR #291**


pkg/client/dagservice/exchange.go
frrist: While not part of this PR, I think we'll need some form of concurrency limiting here

frrist: Also noticing the context passed to GetBlocks isn't used in here, might be easier to swap to an error group.

PR #297, opened 2026-01-19
feat: add open command to view CIDs on storacha.link

gitsofaryan:
## Description
Adds `guppy open <cid>` command for JS client parity. Opens CIDs in the default browser via https://storacha.link.

Implements cross-platform browser opening:
- Windows: `rundll32 url.dll,FileProtocolHandler`
- macOS: `open`
- Linux: `xdg-open`

## Testing
```bash
$ guppy open bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi
Opening https://storacha.link/ipfs/bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi
# Opens in default browser

### DIFF
--- a/cmd/open.go
+++ b/cmd/open.go
@@ -1,0 +1,43 @@
+package cmd
+
+import (
+	"fmt"
+	"os/exec"
+	"runtime"
+
+	"github.com/mitchellh/go-wordwrap"
+	"github.com/spf13/cobra"
+)
+
+var openCmd = &cobra.Command{
+	Use:   "open <cid>",
+	Short: "Open CID on https://storacha.link",
+	Long: wordwrap.WrapString(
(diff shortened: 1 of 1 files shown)


gitsofaryan:
Thanks @alanshaw for the clarification that makes sense. I missed that Guppy is Forge-only and that Forge data isn’t accessible via public gateways.

To make sure future contributions are better aligned: which kinds of commands would you recommend focusing on for Guppy?

For example, would read-only / introspection commands (e.g. account, plan, usage, CID metadata) or workflow commands around delegations/proofs be more useful for Guppy’s audience?

Happy to follow whatever direction makes the most sense.

PR #304, opened 2026-01-22
explore: Demonstrate and test retrieval through Kubo

Peeja:
This doesn't necessarily all need to be merged, although it probably wouldn't hurt to. We do need the changes to the gateway server there to make this possible.

A little breakdown here:
* We need to implement the [Delegated Routing spec](https://specs.ipfs.tech/routing/http-routing-v1/) (or at least the [Content Routing API](https://specs.ipfs.tech/routing/http-routing-v1/#content-routing-api)) to make the connection. We point Kubo at the routing endpoint, and that points all requests to the Gupway gateway.
* The gateway needs to serve all of its content over TLS, or else Kubo will ignore it. It doesn't say anything, it just treats non-TLS HTTP as a protocol it doesn't know and skips it. So, we need to construct a phony cert for `localhost` and tell Kubo not to verify it. That's safe, because we know we control both ends of the conversation, on the same box.
* Unfortunately, that setting not to verify TLS doesn't apply to the _routing_ requests. They'll fail with a made up certificate. _Fortunately_, the routing system is happy to make its requests over plain HTTP if that's where you point it. So, annoyingly, we have to run two separate servers in the gateway process: one for the gateway proper using TLS, and one for the routing API without TLS. The latter is disabled unless you explicitly set a port for it to run on.

### DIFF
--- a/cmd/gateway/serve.go
+++ b/cmd/gateway/serve.go
@@ -40,6 +40,9 @@ import (
 const (
 	// port is the default port to run the gateway on.
 	port = 3000
+	// routingPort is the default port for the delegated routing server (HTTP, no TLS).
+	// Set to 0 to disable.
+	routingPort = 0
 	// blockCacheCapacity defines the default number of blocks to cache in memory.
 	// Blocks are typically <1MB due to IPFS chunking, so an upper bound for how
 	// much memory the cache will utilize is approximately this number multiplied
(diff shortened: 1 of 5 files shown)


cmd/gateway/serve.go
```diff
@@ -252,6 +328,12 @@ func init() {
 	serveCmd.Flags().String("log-level", "", "Logging level for the gateway server (debug, info, warn, error)")
 	cobra.CheckErr(viper.BindPFlag("gateway.log_level", serveCmd.Flags().Lookup("log-level")))
 
+	serveCmd.Flags().String("tls-cert", "", "Path to TLS certificate file (enables HTTPS)")
+	cobra.CheckErr(viper.BindPFlag("gateway.tls-cert", serveCmd.Flags().Lookup("tls-cert")))
+
+	serveCmd.Flags().String("tls-key", "", "Path to TLS key file (enables HTTPS)")
+	cobra.CheckErr(viper.BindPFlag("gateway.tls-key", serveCmd.Flags().Lookup("tls-key")))
```
alanshaw:
```suggestion
	cobra.CheckErr(viper.BindPFlag("gateway.tls_key", serveCmd.Flags().Lookup("tls-key")))
```

pkg/config/gateway.go
```diff
@@ -5,6 +5,13 @@ import "errors"
 type GatewayConfig struct {
 	// Port is the port to run the gateway on.
 	Port int `mapstructure:"port" flag:"port" toml:"port"`
+	// TlsCert is the path to the TLS certificate file. If empty, TLS is disabled.
+	TlsCert string `mapstructure:"tls-cert" flag:"tls-cert" toml:"tls-cert"`
```
alanshaw:
```suggestion
	TlsCert string `mapstructure:"tls_cert" flag:"tls-cert" toml:"tls_cert"`
```

pkg/config/gateway.go
```diff
@@ -5,6 +5,13 @@ import "errors"
...
+	// TlsKey is the path to the TLS key file. If empty, TLS is disabled.
+	TlsKey string `mapstructure:"tls-key" flag:"tls-key" toml:"tls-key"`
```
alanshaw:
```suggestion
	TlsKey string `mapstructure:"tls_key" flag:"tls-key" toml:"tls_key"`
```

PR #319, opened 2026-02-02
feat: Support Postgres DB

### LINKED ISSUE #316: Guppy's prep DB can be in Postgres (maybe secret option)
* The DB config option can point to PG instead of a SQLite file
* Does not need to support any concurrency beyond what SQLite offers, it's purely for safer durability for giant uploads.
* It's okay if it has pitfalls and footguns for now, as long as we can successfully coach Ian/FF/Smithsonian to use it safely. They want a stronger guarantee they won't have to start over before they start; we don't need to support arbitrary customers yet. It can even be a "hidden" option in some form if necessary.

Peeja:
Support using PG as a backend instead of SQLite. Explicitly only for use by a single process at a time. This is only for durability, not coordination.

Closes #316

#### PR Dependency Tree

* **PR #319** 👈

### DIFF
--- a/pkg/preparation/repo.go
+++ b/pkg/preparation/repo.go
@@ -94,6 +94,18 @@ func openPostgresRepo(ctx context.Context, connURL string, opts ...sqlrepo.Optio
 }
 
 func migrate(ctx context.Context, db *sql.DB, dialect sqlrepo.Dialect) error {
+	// Migration 20240101000000 was added later, containing what used to be a
+	// separate base `schema.sql` file, which means it wasn't recorded as a
+	// migration. But any DB which has a `goose_db_version` table effectively has
+	// that migration applied. If it was applied as `schema.sql` and didn't get a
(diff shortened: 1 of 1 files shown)


pkg/preparation/sqlrepo/migrations/20240101000000_init_schema.sql
```diff
@@ -38,31 +25,15 @@ CREATE TABLE IF NOT EXISTS space_sources (
...
-) STRICT;
-
 CREATE TABLE IF NOT EXISTS fs_entries (
   id BLOB PRIMARY KEY,
   source_id BLOB NOT NULL,
   space_did BLOB NOT NULL,
   path TEXT NOT NULL,
   last_modified INTEGER NOT NULL,
-  MODE INTEGER NOT NULL,
+  "mode" INTEGER NOT NULL,
```
alanshaw: Is this a breaking change?

Issue #201, opened 2025-12-01
`guppy space generate` can name a space


Peeja:
1) you should be able to pick a name for the space during generation, and it should be included in the space "facts" like we do with the js client
2) you should be able to see a space's name in the "space list" command
3) don't worry about using the space name as an alias - that will be in a separate ticket

PR #300, opened 2026-01-20
feat: add output key option to space generation command

laciferin2024:
Fixes #178 

```shell
go run . space generate --output-key

WARNING: This is your space private key. Keep it secret and secure!
Space Key (base64):
uG/8avN6yzJYBWi/NFHlNpXN3223+P2JHu1pVbvA2O/9uU/c90PF9+rYm2sgBBAacpEVz/89AnR1FAzZ88/qQmw==
```

PR #322, opened 2026-02-03
feat:  space generate --name 

laciferin2024:
Solves #201

Implemented support for picking a name for a space during generation and displaying it in the space list

- Implemented optional `--name` flag : `generate space --name <name>`
- Introduced the `Space` struct
```
type Space struct {
    DID  did.DID
    Name string
}
```
- Implemented logic to extract the name fact from delegations and populate the newly created `Space` struct.
- Verified the logic with `space list` and `space list --json`


laciferin2024:
@Peeja @alanshaw  Please review post merge of #300 .

laciferin2024:
Relates #331

PR #339, opened 2026-02-12
feat: Show cause on context cancel

Peeja:
I'm tired of seeing log messages like `failed to add blob Shard[id=1cb59615-1d52-4821-a260-2b2840c043a7]: failed to get reader for blob Shard[id=1cb59615-1d52-4821-a260-2b2840c043a7]: iterating nodes in shard b0dc805a-7c2b-40db-9d4c-a4d869f66718: failed to get sizes of blocks in shard b0dc805a-7c2b-40db-9d4c-a4d869f66718: context canceled` and not knowing what actually canceled the context. Unfortunately, Go doesn't currently make it terribly easy to see that. This pattern should make it more obvious what's happening.

1. Rather than use `ctx.Err()` in our error messages, use `ctxutil.CausedError()`. It's similar to `context.Cause()`, except that simply returns the cause directly. This returns a wrapped error which shows the `ctx.Err()` and that it was *caused* by the `context.Cause()`.

2. Rather than use errors coming back from library functions directly in our wrapped errors, use `ctxutil.ErrorWithCause()`. This accounts for library functions which themselves return the `ctx.Err()` rather than noticing the `context.Cause()` (likely because they were written before `context.Cause()` existed). It checks to see if the returned error was the context's `ctx.Err()`, and if so, and if there's a `context.Cause()` available, wraps them like `ctxutil.CausedError()` for readability.

This means touching *a lot* of call sites, but with a very clear pattern, and hopefully valuable results.

#### PR Dependency Tree

* **PR #339** 👈
  * **PR #344**
    * **PR #358**
      * **PR #359**
        * **PR #360**
          * **PR #361**
            * **PR #362**
              * **PR #363**
                * **PR #364**
                  * **PR #365**
                    * **PR #366**

### DIFF
--- a/internal/ctxutil/ctxutil.go
+++ b/internal/ctxutil/ctxutil.go
@@ -31,14 +31,14 @@ func Cause(ctx context.Context) error {
 	return fmt.Errorf("%w, cause: %w", err, cause)
 }
 
-// ErrorWithCause takes an error which should be returned from a function that
+// EnrichWithCause takes an error which should be returned from a function that
 // also takes a context, and the context itself. If the context was canceled or
 // its deadline was exceeded *with a cause*, and the error returned was the
 // `ctx.Err()` without including the cause, then this function will return a new
 // error that wraps both. This is an ergonomic convenience to adapt functions
 // which respond to `ctx.Done()` but don't include the cause in their returned
 // error, likely because they were written before context causes were added.
-func ErrorWithCause(err error, ctx context.Context) error {
+func EnrichWithCause(err error, ctx context.Context) error {
 	cause := context.Cause(ctx)
 	if cause != nil && errors.Is(err, ctx.Err()) && !errors.Is(err, cause) {
 		return fmt.Errorf("%w, cause: %w", err, cause)
--- a/internal/ctxutil/ctxutil_test.go
+++ b/internal/ctxutil/ctxutil_test.go
@@ -62,21 +62,21 @@ func TestCausedError(t *testing.T) {
 func TestErrorWithCause(t *testing.T) {
 	t.Run("returns nil when err is nil and context is not done", func(t *testing.T) {
 		ctx := t.Context()
-		assert.Nil(t, ctxutil.ErrorWithCause(nil, ctx))
+		assert.Nil(t, ctxutil.EnrichWithCause(nil, ctx))
 	})
 
 	t.Run("returns nil when err is nil and context is canceled with cause", func(t *testing.T) {
 		cause := errors.New("shutting down")
 		ctx, cancel := context.WithCancelCause(t.Context())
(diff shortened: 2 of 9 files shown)


internal/ctxutil/ctxutil.go
volmedo: non-blocking suggestion: I don't love the names of the functions, they sound quite similar and difficult to tell which does what. But I get they are difficult to name, what about something like?
- just `Err` for `CausedError`: the function is to be used in the same places as the usual `ctx.Err()` so the intent is clear. Calling it as `ctxutil.Err(ctx)` still gives information that this is a different call.
- `EnrichWithCause` or just `Enrich` for `ErrorWithCause`: this is what the function does in the end. It's not a new error, it's the same error with additional info if available.

volmedo (approved):
I like this ❤️ 

Question: I guess this needs to be complemented with actually adding causes when contexts are cancelled/timed out. Is that done in a different PR?

internal/ctxutil/ctxutil_test.go
```diff
@@ -0,0 +1,149 @@
...
+		<-ctx.Done()
+		err := ctxutil.CausedError(ctx)
+		assert.ErrorIs(t, err, context.DeadlineExceeded)
+		assert.Equal(t, "context deadline exceeded", err.Error())
+	})
+
+	t.Run("returns error wrapping both deadline exceeded and cause", func(t *testing.T) {
+		cause := errors.New("query too slow")
+		ctx, cancel := context.WithCancelCause(context.Background())
+		cancel(cause)
```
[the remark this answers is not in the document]
Peeja: Whoops! Yes, thanks!

internal/ctxutil/ctxutil.go
Peeja: Yeah, I like that. I wasn't a big fan of these names either.

internal/ctxutil/ctxutil.go
```diff
@@ -0,0 +1,47 @@
...
+//
+//   - nil, if the context is not done
+//   - the error from `ctx.Err()` if there is no separate cause, which will be
+//     either [context.Canceled] or [context.DeadlineExceeded]
+//   - an error stating that the [context.Canceled] or
+//     [context.DeadlineExceeded] was caused by the cause, if there is a cause
+//
+// For a context with a cause, the returned error will wrap both the error and
+// the cause, so you can use `errors.Is`, `errors.As`, and `errors.Unwrap`.
+func CausedError(ctx context.Context) error {
```
alanshaw: I would perhaps just call this `Cause` and frame it as a better version of `context.Cause(context.Context) error` that includes the cancellation error in the wrapping.

alanshaw (approved):

PR #341, opened 2026-02-16
feat: preparation check

### LINKED ISSUE #311: Preparation Check
#### Potential upload failure scenarios during Shutdown

Upload operations can encounter data inconsistencies when shutdowns occur at critical points in the process (especially in older versions)

1. **Incomplete DAG Scanning**
   - During filesystem scanning, an `FSEntry` may be created before a `DAGScan` is generated
   - If shutdown occurs between these steps, the result is no scan record and no node creation, causing the filesystem entry to be silently ignored

2. **Partial UnixFS Node Creation**
   - A UnixFS node can be created but shutdown may occur before its links are established
   - This produces a node that appears to have no files or links
   - The node will have an incorrect CID for the intended content, but will *not* fail `NodeReader` CID validation since the CID is technically correct for a node with zero links

3. **Orphaned Nodes**
   - A node can be created but fail to be added to a shard, leaving it orphaned in the database

### Proposed Preparation Check

The preparation check aims to: (1) detect the above integrity issues, (2) check for completeness and (3) put the database in a state where re-running the upload will trigger repair actions and complete the upload

#### Upload Completeness Check

Insure the upload has both a root FSEntry and a root cid

#### File System Integrity Check

Starting with an upload's root `FSEntry`:

**For file entries:**
- Ensure a `DAGScan` exists
  - If missing, create one to trigger scanning on upload
  - If present, inspect the DAG starting with the root node:
    - If root is a block, continue
    - If root is not a block, ensure it has children
    - Recursively inspect each child node
    - If any UnixFS node lacks children when expected, the DAG is invalid
      - Delete the node CID from the `DAGScan`
      - This makes the scan incomplete, triggering a rescan

**For directory entries:**
- Recursively inspect all child `FSEntry` objects
- Ensure a `DAGScan` exists
  - If missing, create one to trigger scanning on upload
  - If present:
    - If any child `FSEntry` had errors, mark directory DAG as invalid
    - If no child errors, scan the DAG from root CID to ensure directory entries match (may require UnixFS iterator for HAMT directories)
    - If directory DAG is invalid:
      - Delete the node CID from the `DAGScan`
      - This makes the scan incomplete, triggering a rescan

If the root directory DAG was invalid, remove the root CID from the upload.

#### Node Integrity Check

For every node associated with the upload:
- Insure it has a node_upload record
- If missing, create one with no `shard_id`

#### Node Completeness Check

Check no nodes are returned by NodesNotInShards

#### Shard Completeness Check

Check every shard associated with the uplaod is in the BlobStateAdded state

Check no shards are returned by ShardsNotInIndexes

#### Index Completeness Check

Check every index associated with the upload is in a BlobStateAdded state

### User Notification

If any integrity or completeness issues are detected, or any completeness check inform the user and recommend re-running the upload.

volmedo:
closes #311 

Implements the `guppy upload check` command, which performs checks on the state of uploads in the database and attempts to fix uploads in a bad state so that they can be re-run.

Currently, the semantics are "check, and optionally try to repair" with the `--repair` flag. Alternatively, attempting to repair could be the default, maybe changing the command to `guppy upload repair`, with a `--dry-run` flag to just output a report of issues. I don't have a strong opinion, thoughts welcome.

I also added a script to create a test database with uploads in different bad states (./test/baduploads) to allow manual testing of the command without interfering with the main database.

`checkFileSystemIntegrity` is the most complex check, so eyes on that one are welcome.

### DIFF
--- a/cmd/upload/check/check.go
+++ b/cmd/upload/check/check.go
@@ -63,7 +63,7 @@ Specify both space and source to check a specific upload.`,
 		}
 		defer repo.Close()
 
-		client := cmdutil.MustGetClient(cfg.Repo.Dir)
+		client := cmdutil.MustGetClient(cfg.Repo.Dir, cfg.Network)
 
 		// Determine which uploads to check
 		var uploadsToCheck []uploadInfo
--- a/cmd/account/list.go
+++ b/cmd/account/list.go
@@ -31,7 +31,7 @@ var listCmd = &cobra.Command{
 		if err != nil {
 			return err
 		}
-		c := cmdutil.MustGetClient(cfg.Repo.Dir)
+		c := cmdutil.MustGetClient(cfg.Repo.Dir, cfg.Network)
 
 		accounts, err := c.Accounts()
 		if err != nil {
--- a/cmd/blob/ls.go
+++ b/cmd/blob/ls.go
@@ -1,0 +1,100 @@
+package blob
+
+import (
+	"bytes"
+
+	"github.com/dustin/go-humanize"
+	"github.com/ipld/go-ipld-prime/codec/dagjson"
+	"github.com/mitchellh/go-wordwrap"
+	"github.com/spf13/cobra"
+	spaceblobcap "github.com/storacha/go-libstoracha/capabilities/space/blob"
+	"github.com/storacha/go-libstoracha/digestutil"
+	"github.com/storacha/go-ucanto/core/delegation"
+
+	"github.com/storacha/guppy/internal/cmdutil"
+	"github.com/storacha/guppy/pkg/client"
+	"github.com/storacha/guppy/pkg/config"
+)
+
+const pageSize = 1000
+
+var lsFlags struct {
+	proofsPath string
+	long       bool
+	human      bool
+	json       bool
+}
+
+func init() {
+	lsCmd.Flags().StringVar(&lsFlags.proofsPath, "proof", "", "Path to archive (CAR) containing UCAN proofs for this operation.")
+	lsCmd.Flags().BoolVarP(&lsFlags.long, "long", "l", false, "Display detailed information about blobs.")
+	lsCmd.Flags().BoolVarP(&lsFlags.human, "human", "H", false, "Display blob sizes in human-readable format (only applicable when used with --long).")
+	lsCmd.Flags().BoolVar(&lsFlags.json, "json", false, "Output as newline delimited JSON.")
+}
+
+var lsCmd = &cobra.Command{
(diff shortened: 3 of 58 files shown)


cmd/upload/check/check.go
```diff
@@ -0,0 +1,385 @@
...
+			"2. File System Integrity Check - Validates FS structure and DAG integrity\n"+
+			"3. Node Integrity Check - Ensures all nodes have upload records\n"+
+			"4. Node Completeness Check - Verifies all nodes are in shards\n"+
+			"5. Shard Completeness Check - Verifies all shards are uploaded and indexed\n"+
+			"6. Index Completeness Check - Verifies all indexes are uploaded\n\n"+
+			"By default, runs in dry-run mode (reports issues without fixing). "+
+			"Use --repair to automatically fix issues where possible.\n\n"+
+			"If no arguments are provided, checks all uploads. "+
+			"Specify a space to check only uploads for that space. "+
+			"Specify both space and source to check a specific upload.",
```
frrist: using backticks (` `) here would be cleaner, could remove the manual new lines and such.

pkg/preparation/sqlrepo/dags.go
```diff
@@ -87,6 +87,26 @@ func (r *Repo) dagScanScanner(sqlScanner sqlScanner) model.DAGScanScanner {
...
+		FROM dag_scans
+		WHERE fs_entry_id = ?
+	`)
+	if err != nil {
+		return nil, fmt.Errorf("failed to prepare statement: %w", err)
+	}
+	row := stmt.QueryRowContext(ctx, fsEntryID)
+	ds, err := model.ReadDAGScanFromDatabase(r.dagScanScanner(row))
+	if errors.Is(err, sql.ErrNoRows) {
+		return nil, nil // Not found
```
frrist: Blocking: `nil, nil` is a footgun. Every caller has to remember to check for a nil result _and_ a nil error, which is easy to forget and leads to nil pointer dereferences down the line. Would strongly prefer to return a package specific not found typed error here.

volmedo: I agree that returning a package specific error may be clearer about the intent of the code, but aside from that I don't think it's so advantageous over returning `nil, nil`. Callers need to check for a nil error, as usual, and then handle the result, with that being checking for a nil result or a specific error. If a specific error is returned and callers forget to check for it they'll hit the same nil pointer dereference issue.

Also, this code is consistent with how this situation is handled in other parts of the `sqlrepo` package.

PR #345, opened 2026-02-18
chore: fix flaky tests relying on map iteration order

volmedo:
Fix a couple tests that showed flakiness in CI runs, plus another one I caught when confirming the fixes were working as expected.

### DIFF
--- a/pkg/verification/retrieval_test.go
+++ b/pkg/verification/retrieval_test.go
@@ -13,6 +13,7 @@ import (
 	"github.com/storacha/go-libstoracha/blobindex"
 	assertcap "github.com/storacha/go-libstoracha/capabilities/assert"
 	captypes "github.com/storacha/go-libstoracha/capabilities/types"
+	"github.com/storacha/go-libstoracha/digestutil"
 	"github.com/storacha/go-libstoracha/testutil"
 	ed25519signer "github.com/storacha/go-ucanto/principal/ed25519/signer"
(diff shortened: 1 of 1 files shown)


pkg/verification/retrieval_test.go
```diff
@@ -72,15 +72,21 @@ func TestStatBlocks(t *testing.T) {
...
+				require.Equal(t, uint64(len(block1Data)), stat.Size)
+				require.Empty(t, stat.Links, "raw blocks have no links")
+			case block2Hash.String():
+				require.Equal(t, uint64(cid.Raw), stat.Codec)
+				require.Equal(t, uint64(len(block2Data)), stat.Size)
+				require.Empty(t, stat.Links, "raw blocks have no links")
+			default:
+				t.Errorf("unexpected block stat with digest %s", stat.Digest.String())
+			}
+		}
```
Peeja: **suggestion:** We already iterated to get `stats`, and I think we just do that to assert the length is 2, which is good, but doesn't actually prove we have both of these blocks, since they could be the same. What about a single iteration: replace `range stats` here with `range verification.StatBlocks(…)`, and also collect a slice of the hashes, and then `require.ElementsMatch()` them? I wouldn't make a big deal out of making sure it's exhaustive, but I think that may come out cleaner anyhow.

volmedo: good catch, thanks!

PR #354, opened 2026-02-23
feat: Use single DB query to iterate for index

Peeja:
#### PR Dependency Tree

* **PR #354** 👈

### DIFF
--- a/pkg/preparation/sqlrepo/indexes.go
+++ b/pkg/preparation/sqlrepo/indexes.go
@@ -348,3 +348,41 @@ func (r *Repo) ShardsForIndex(ctx context.Context, indexID id.IndexID) ([]*model
 
 	return shards, nil
 }
+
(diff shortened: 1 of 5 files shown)


pkg/preparation/sqlrepo/indexes.go
```diff
@@ -348,3 +348,41 @@ func (r *Repo) ShardsForIndex(ctx context.Context, indexID id.IndexID) ([]*model
 
 	return shards, nil
 }
+
+func (r *Repo) ForEachNodeInIndex(ctx context.Context, indexID id.IndexID, yield func(shardDigest multihash.Multihash, nodeCID cid.Cid, nodeSize uint64, shardOffset uint64) error) error {
```
volmedo: nit: what do you think of using the idiomatic iterator syntax?

We could create a new `NodeInIndex` type as:
```
type NodeInIndex struct {
    Cid                cid.Cid
    Size               uint64
    ShardDigest multihash.Multihash
    ShardOffset uint64
```
and then have this function be:
```
func (r *Repo) NodesInIndex(ctx context.Context, indexID id.IndexID) iter.Seq2[NodeInIndex, error]
```
and then the caller can range over the results and do whatever is needed with them.

PR #400, opened 2026-04-10
feat: configure replication count

alanshaw:
Allows the replication count to be configured. It's a hidden flag, since we don't really want users to change this, it's mostly for testing where we might want/need to reduce it to 1.

Aside, we should probably allow a couple of the other preparation flags to be configured in TOML also.


pkg/config/preparation.go
```diff
@@ -0,0 +1,17 @@
+package config
+
+import "fmt"
+
+type PreparationConfig struct {
```
frrist: I'm not keen on the naming of this, as a user reading the config `Preparation` will throw me off. I _think_ I'd prefer `Upload` or something along those lines, parallelism will likely eventually be included there too, later maybe even block_size and/or dag_type (balanced vs trickle). But this isn't a blocking comment, up to you.

frrist (approved):

pkg/config/preparation.go
alanshaw: Yes same. I will change to "upload".

PR #403, opened 2026-04-16
cd: align container build and publish process with the storacha RFC

frrist:
- Moves container publishing out of goreleaser into a dedicated `Container`
workflow. Adds `main` and `sha-<short>` tags on push to main, a PR
build-check, and RFC-aligned Dockerfile (alpine:latest runtime, USER
nobody, `--platform=$BUILDPLATFORM` cross-compilation, no RUN in the
runtime stage so QEMU is not required).

- Adds ghcr.io/storacha/guppy:main-dev and :sha-<short>-dev images so
developers can pull a guppy container that mirrors prod but ships with
delve, randdir, and diagnostic tools for debugging. Extends the PR
build-check to validate Dockerfile.dev too, and fixes a latent
cross-compile bug where `go install` placed delve/randdir under
/go/bin/linux_<arch>/ but the COPY targeted /go/bin/ — arm64 builds of
the dev image would have failed. QEMU is set up only for the dev job
since the prod runtime stage still has no RUN commands.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,15 @@
+# dirs
+.git
+.github
+.claude
+.idea
+docs
+
+# files
+.env*
+*.md
+.goreleaser.yaml
+docker-compose*.yml
+LICENSE
+LICENSE.md
+version.json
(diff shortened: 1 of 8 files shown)


frrist:
Given this is more or less a port of the container release process used by piri, indexing-service, etc. I am going to land without review as I am tired of building containers locally. If this breaks main wil fix ofc.

Peeja:
Awesome! Sorry for the post-facto, but 👍🏻

PR #408, opened 2026-04-17
feat: Better parallelism

### LINKED ISSUE #267: Improve parallelism in Storacha workers
# What

Currently, the Storacha upload and post process functions operate with parallelism that works like so:

---

read closed blobs -> array of closed blobs
spin up N parallel processors and iterate through the blobs
wait for all the blobs to finish

---

and then seperately, loop again as there are new closed blobs.

The problem occurs when one of the uploads towards the end is particularly slow
let's say I read 5 closed blobs, then iterate through them
4 have finished, while 1 is taking a while to finished
now I have 10 more new closed blobs since I first read that are essentially idle, and the whole thing is stuck till the slow upload finishes

---

We need something where as each worker finishes, we refresh the list of available jobs and keep iterating.

Peeja:
Previously, the `Worker` would always run in batches. When signaled, it would find new work to do, then perform all of that work, then wait for a new signal. Some workers would perform their work in parallel, but only *within* the batch, meaning that one long-lived task in the batch would waste the rest of the parallelism, and no new tasks would be pulled in until the entire batch completed.

Now, `Worker` understands tasks and parallelism natively. Whenever it has free parallelism, it looks up pending tasks. Then it queues those as it works through them. Notably, each `Worker` definition is responsible for tracking what's in flight and not re-queuing it. That is, since we don't store that we've placed a task in the queue, we'll keep finding it in the DB until the task is complete, and we need to filter that out. But it's pretty straightforward to manage that by DB ID. We keep that in memory rather than in the DB, because if the process is interrupted it naturally results in a consistent state: the queue is emptied, and so is the set of what we consider "in-flight".

My gut says that there's further we can push this, but this is a significant improvement as it is. I've tested it with `doupload`. #409 demonstrates it with the (shell version of the) Smelt-based `doupload`.

Closes #267


pkg/preparation/storacha/storacha.go
```diff
@@ -58,164 +56,123 @@ type ReaderForIndexFunc func(ctx context.Context, indexID id.IndexID) (io.ReadCl
...
+	tasks := make([]gtypes.IDTask, 0, len(closedShards))
+	for _, shard := range closedShards {
+		tasks = append(tasks, gtypes.IDTask{
+			ID: shard.ID(),
+			Run: func(ctx context.Context) (error, error) {
+				if err := a.addBlob(ctx, shard, spaceDID); err != nil {
+					err = fmt.Errorf("failed to add shard %s: %w", shard, err)
+					// [gtypes.BlobUploadError]s are non-fatal.
+					var errBlobUpload gtypes.BlobUploadError
+					if errors.As(err, &errBlobUpload) {
```
alanshaw, line 97: Could just use `errors.Is`? You're not using `errBlobUpload` at all.

pkg/preparation/types/errors.go
```diff
@@ -190,9 +190,14 @@ func (e BlobUploadErrors) Error() string {
...
+	return e.errs
+}
+
+// IDTask represents a task that can be identified and deduplicated by an
+// [id.ID]. The task's Run function returns two errors: a non-fatal error that
+// can be collected and reported after all tasks have completed, and a fatal
+// error that should cause immediate cancellation of all other tasks.
+type IDTask struct {
+	ID  id.ID
+	Run func(context.Context) (error, error)
```
alanshaw: Perhaps use a type for the non-fatal error e.g.

```go
type Report struct {
  Warnings []error
}
```

It's just weird IMO to see a function that returns two errors.

pkg/preparation/storacha/storacha.go
alanshaw: Minor, but personally I'd flip this round to return early for the fatal case.

pkg/preparation/storacha/storacha.go
```diff
@@ -306,7 +263,7 @@ func (a API) addBlob(ctx context.Context, blob model.Blob, spaceDID did.DID) err
 		}
 	}
 
-	if err := a.updateBlob(ctx, blob); err != nil {
+	if err := a.updateBlob(context.WithoutCancel(ctx), blob); err != nil {
```
alanshaw, line 267: Why do we want this?

pkg/preparation/types/errors.go
```diff
@@ -190,9 +190,14 @@ func (e BlobUploadErrors) Error() string {
...
-	}
-	return errs
+	return e.errs
+}
+
+// IDTask represents a task that can be identified and deduplicated by an
+// [id.ID]. The task's Run function returns two errors: a non-fatal error that
+// can be collected and reported after all tasks have completed, and a fatal
+// error that should cause immediate cancellation of all other tasks.
+type IDTask struct {
```
alanshaw: Do (should?) we have non-ID tasks? Is this just a `Task`?

pkg/preparation/types/errors.go
```diff
@@ -190,9 +190,14 @@ func (e BlobUploadErrors) Error() string {
...
+	ID  id.ID
```
alanshaw: Suggestion: add a non-unique `Desc string` which could be used in error messages.

pkg/preparation/types/errors.go
alanshaw: Should this live in `errors.go`?

alanshaw (approved):
Happy for this to merge but please consider the feedback. I would most like to have a type for the non-fatal task return value - I suggested a "report" that could contain warning messages, but I'm not attached to that naming.

pkg/preparation/types/errors.go
Peeja: My intent here was separation of concerns: the `Worker` code doesn't handle deduping tasks and leaves it entirely to the worker definition which uses `Worker`. In particular, that means the non-parallel workers don't have to make up a bogus "ID" to satisfy an interface, they just don't do any deduping. I'm certainly open to thoughts here, though; I debated it a bit myself.

pkg/preparation/types/errors.go
Peeja: Oooh, I like that.

pkg/preparation/storacha/storacha.go
Peeja: Because we're recording something that happened externally. We need to make sure we do that. We can't make the DB update truly transactional with the external side effect, but we can at least make sure we don't give up on recording what happened because the context canceled. Otherwise we might do things like upload the same shard twice. (And given `updateBlob()` is a simple DB update, the potential cost is low.)

pkg/preparation/storacha/storacha.go
Peeja: > Could just use `errors.Is`? You're not using `errBlobUpload` at all.

`errors.Is` is for instances, `errors.As` is for types, unless I'm misunderstanding.

Peeja:
@alanshaw @hannahhoward How's this look? Part of the intention with this `TaskError` model is that we can use it a level up as well to track the non-fatals across the workers a bit more easily, but I've spent enough time on this thing as it is, so I'd rather leave that for the future.

